PR #36, opened 2026-02-21
Improved CellComputationAutogradHooks by may_match_twice predicate

### LINKED ISSUE #3: Identify unmatched autograd saved tensors pack args with new/old training replay cache variant
**Is your feature request related to a problem? Please describe.**

The pack/unpack mechanism in `autograd_hooks.py` saves memory in the gradient computation. At the moment, a certain number of tensors of tracked shapes are not matched. This leads to more GPU memory being used than necessary.

**Describe the solution you'd like**

The old training replay cache uses own SDPA operators in `sdpa_op.py`, which compute `backward` naively. Even for these, pack args are not matched. Find out why!

For the new training replay cache, we use the `backward` of PyTorch SDPA with zero-padded queries. Many more pack args are not matched. For improving this, may have to study the `backward` implementation of the PyTorch SDPA.

For `flex_attention`, things may be simpler. If this is faster than zero-padded queries, should maybe retire the latter? Need to find out what these unmatched pack args are for `flex_attention`.

**Describe alternatives you've considered**

Ideally, we'd have a fused operator (like in `sdpa_op.py`) with a non-naive `backward`. This needs low-level implementation (CUDA or Triton). Apart from being faster that PyTorch SDPA (no zero-padding) and requiring less memory (fusion with KV cache update), we then know exactly what intermediates are created.

mseeger:
* Reduce number of comparisons in `CellComputationAutogradHooks` by removing annotations once matched once or twice. Can be controlled by `may_match_twice` predicate
* Extra mechanism `track_unmatched_annotations` to monitor what unmatched pack args are matched against

Closes #3.

----

### DIFF
--- a/keys_values/finetune/longcontext_full.py
+++ b/keys_values/finetune/longcontext_full.py
@@ -891,7 +891,9 @@ def wrap_gpt_model(
             may_match_twice=may_match_twice_factory(grad, gpt_model),
         )
         if grad.max_match_trials_pack_arg is not None:
-            autograd_hooks_kwargs["max_match_trials_pack_arg"] = grad.max_match_trials_pack_arg
+            autograd_hooks_kwargs["max_match_trials_pack_arg"] = (
+                grad.max_match_trials_pack_arg
+            )
         if cpu_offload_device is not None:
             common_kwargs["head_model"] = head_model.to(device=cpu_offload_device)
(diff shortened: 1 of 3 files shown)


mseeger:
TODO: Change `test_gradient`, `test_gradient_main` to incorporate this as well.

PR #47, opened 2026-03-03
CPU offloading of (quantized) KV cache buffers (#47)

### LINKED ISSUE #18: Implement CPU offloading for KV cache buffers
**Is your feature request related to a problem? Please describe.**

We already support quantized KV cache buffers, but in GPU memory, as well as KV cache checkpointing during `backward`. We'd like to offload KV cache buffers to CPU for layers not worked on right now. This saves memory during inference and allows for larger KV cache lengths. `vLLM` is doing this as well.

With this feature, we can run inference with 4B models on a single P4 instance, and fine-tuning 4B models together with CPU offloading during backward pass (which is implemented already).

**Describe the solution you'd like**

This should not be very hard. We already support KV cache checkpointing to CPU. Just need to make this available as a KV cache buffer strategy, so it can be used during `forward` as well.

mseeger:
Enables CPU offloading of quantized KV cache buffers, which allows to run inference for bigger models.

This is done by sharing quantizers across all quantized KV cache buffers, but storing the quantizer states per layer to CPU.

Closes #18.

----

### DIFF
--- a/keys_values/finetune/args.py
+++ b/keys_values/finetune/args.py
@@ -16,7 +16,8 @@ from typing import Optional, Tuple, Dict, Any, List
 
 from litgpt.args import EvalArgs as _EvalArgs
 
-from keys_values.kvcache.factory import KVCacheFactory, split_name
+from keys_values.kvcache.factory import KVCacheFactory
+from keys_values.kvcache.consts import split_name
 from keys_values.utils import VerbosityLevels
 
 
@@ -66,7 +67,9 @@ class KVCacheArgs:
             this number of last recently inserted tokens are not evicted.
         init_grace_tokens: Only for `lastrec` cache policy. KV information for
             the first `init_grace_tokens` is not evicted.
-
+        cpu_offload: If `True`, KV cache buffers are offloaded to CPU during the
+            forward pass. At the moment, this is implemented only for quantized
+            KV cache buffers.
     """
 
     name: str
@@ -77,6 +80,7 @@ class KVCacheArgs:
     allocate_buffers: bool = False
     grace_period: int = 0
     init_grace_tokens: int = 0
+    cpu_offload: bool = False
     # Legacy (these are global args now)
     verbose: Optional[str] = None
     attention_forward_temp_size_gb: Optional[float] = None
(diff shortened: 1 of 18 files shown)


mseeger:
* Forward options:

[1] cell, then layer, then chunk/cell (we do this right now)
[2] layer, then cell, then chunk/cell

Assume: Mem for KV cache = Mem for cell embeddings

How many copies CPU/GPU?
[1]
- embeddings: -
- caches: n_cells * n_layers

[2]
- embeddings: n_cells * n_layers
- caches: n_layers

==> Pretty much the same. Stick with [1]

Another way to look at it (if quantization is used):

[1] Embeddings don't need quant, but caches n_cells * n_layers
[2] Caches don't need quant along cells, but embeddings do

==> Easiest to stick with [1]

* What we have:
- KV cache buffers (quantized):
  - Each has its own quantizers with buffers on GPU
  - Dequantization buffers shared
  - Quantizer.create_quantizer_state -> what we need!
- KV cache buffers (not quantized):
  - Each has its own buffers on GPU
- KV cache checkpointing
  - Stores KV cache buffers for certain chunk_ids; only in single layer
  - I/O via DefaultKVCacheBuffers -> not so useful

OK: Do not use code in `checkpoints.py`!

* What is missing:
- Quantizers can be shared by several KVCacheBuffer's
- QuantizerState owns the copying via
  `copy_` (buffer -> state), `restore` (state -> buffer)
- Container to store the quantizer states on CPU

* Design

Current:
- QuantizedKVCacheBuffers:
  - quantizer_k -> Quantizer
  - quantizer_v -> Quantizer
- Quantizer:
  - No reference back to QuantizedKVCacheBuffers. No idea when
    buffer is changing!
  - create_quantizer_state -> QuantizerState
- QuantizerState:
  - quantizer -> Quantizer (reference back)
  - copy_ and restore
- N layers:
  N times independent (QuantizedKVCacheBuffers, Quantizer)

What we need:
- N (QuantizedKVCacheBuffers, QuantizerState) pairs, but only 1
  Quantizer
- buffer -> quantizer: Quantizer has the right state

Two possibilities:
- Push: Buffer initializes the state update for quantizer
  (+) Simple; quantizer code can stay the same
  (+) Could also include push to write state back
  (-) Need to remember to do it
- Pull: Quantizer notices buffer changing, initializes state update
  (+) Safer; no need to remember
  (-) More complex; quantizer has to store back reference; dependence?
  (-) State for previous buffer written back only upon change; safe?

Like to have "pull", but without complex code change in quantizer!

Idea: Callback
- Quantizer gets optional `callback(new_buffers, current_buffers)`; it
  also stores `current_buffers` (need not be typed, to avoid bidirectional
  dep.)
- Quantizer methods called from QuantizedKVCacheBuffers gets optional
  `buffers` arg. This is required if callback is given. Before anything
  is done in the method: Call callback, and then store new `buffers`
- All complexity in class maintaining the quantization states:
  - Maintains quantization states
  - Creation of QuantizedKVCacheBuffers's and Quantizer's
  - Definition of callbacks
  - Method `flush`: Quantizer writes back its current state. This would
    be called at the end of an inference pass, to be safe

PR #83, opened 2026-04-08
Add FlashInfer CUDA kernels + Triton score-sum for H2O attention weights

vihangp:
## Summary
                                                                                                                                                                                                                                                                                                                                                                                                                                      
  Adds FlashInfer CUDA kernels (prefill + decode SDPA) and a Triton score-sum kernel to compute attention weight sums for H2O KV cache eviction, replacing the double FlexAttention approach from PR #78.                                                                                                                                                                                                                             
  
  ## How it works                                                                                                                                                                                                                                                                                                                                                                                                                     
                  
  The double FlexAttention baseline calls `flex_attention` twice per layer: once for the standard SDPA output + LSE, and again with reversed Q/K and specially constructed values to extract attention weight sums.                                                                                                                                                                                                                   
  
  Our approach computes attention weight sums in a single forward pass:                                                                                                                                                                                                                                                                                                                                                               
                  
  1. **FlashInfer SDPA** (vendored CUDA kernels): Computes attention output `O` and log-sum-exp `LSE` in a single call (prefill + decode variants)                                                                                                                                                                                                                                                                                    
  2. **Triton score-sum kernel**: Using the LSE from step 1, computes the H2O attention weight sums directly:
     W[kv_head, k] = Σ_q Σ_h exp2(Q·K·scale·log2e - LSE_log2)                                                                                                                                                                                                                                                                                                                                                                         
  with causal masking and GQA support. This avoids the second FlexAttention call entirely.                                                                                                                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                                                                                                                                                      
  ## Environment                                                                                                                                                                                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                                                                                                                                                      
  | | Baseline (main) | FlashInfer (this branch) |                                                                                                                                                                                                                                                                                                                                                                                    
  |---|---|---|                                                                                                                                                                                                                                                                                                                                                                                                                       
  | Branch | `origin/main` (commit `a933349`) | `flashinfer_integration` |
  | GPU | NVIDIA A100-SXM4-40GB | NVIDIA A100-SXM4-40GB |                                                                                                                                                                                                                                                                                                                                                                             
  | PyTorch | 2.9.1+cu126 | 2.9.1+cu126 |                                                                                                                                                                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                                                                                                                                                                      
  ## Results (Full Training: Forward + Backward)                                                                                                                                                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                                                                                                                                                      
  ### Training Loss                                                                                                                                                                                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                                                                                                                                      
  | Step | Baseline (PT 2.11) | Baseline (PT 2.9) | FlashInfer + Triton (PT 2.9) |
  |------|-------------------|-------------------|-------------------|                                                                                                                                                                                                                                                                                                                                                                
  | 1 | 44.250 | 44.500 | 21.250 |                                    
  | 2 | 14.625 | 15.000 | 1.242 |                                                                                                                                                                                                                                                                                                                                                                                                     
  | 3 | 20.500 | 20.375 | 8.438 |                                                                                                                                                                                                                                                                                                                                                                                                     
  | 4 | 19.625 | 20.125 | 3.328 |                                                                                                                                                                                                                                                                                                                                                                                                     
  | 5 | 14.625 | 14.750 | 3.562 |                                                                                                                                                                                                                                                                                                                                                                                                     
  | 6 | 18.750 | 17.125 | 5.438 |                                                                                                                                                                                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                                                                                                                                                                      
  ### Final Validation (forward-only)                                                                                                                                                                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                                                                                                                                      
  | | Baseline (PT 2.11) | Baseline (PT 2.9) | FlashInfer + Triton (PT 2.9) |                                                                                                                                                                                                                                                                                                                                                         
  |---|---|---|---|                                                                                                                                                                                                                                                                                                                                                                                                                   
  | Val loss | 18.160 | 18.106 | 3.657 |
  | Val ppl | 77,032,103 | 72,974,691 | 38.8 |                                                                                                                                                                                                                                                                                                                                                                                        
  | Val time | 1289.5s | 1277.1s | 1098.2s |                                                                                                                                                                                                                                                                                                                                                                                          
  | Peak GPU mem | 26.60 GB | 26.66 GB | 26.66 GB |                                                                                                                                                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                                                                                                                                      
  ### Key Findings                                                                                                                                                                                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                                                                                                                                                                      
  - **Loss difference is NOT from PyTorch version**: Baseline losses are nearly identical on PT 2.11 (18.160) and PT 2.9 (18.106)                                                                                                                                                                                                                                                                                                     
  - **Validation speedup**: 14% faster (1277s → 1098s, same PyTorch version)                                                                                                                                                                                                                                                                                                                                                          
  - **Val loss**: 18.1 → 3.7 (4.95x better)                                 
  - **Val perplexity**: ~73M → 38.8                                                                                                                                                                                                                                                                                                                                                                                                   
  - **GPU memory**: Identical (~26.6 GB)                                                                                                                                                                                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                                                                                                                                                                                                      
  ## Attention Weight Comparison                                                                                                                                                                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                                                                                                                                                      
  Direct comparison using identical Q, K, V inputs (`test_compare_weights.py`):                                                                                                                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                                                                                                                                      
  | | Triton vs Reference | FlexAttention vs Reference |
  |---|---|---|                                                                                                                                                                                                                                                                                                                                                                                                                       
  | Max abs diff | 0.000001 | 1.374193 |
  | Mean abs diff | 0.000000 | 0.750020 |                                                                                                                                                                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                                                                                                                                                                      
  **Key finding**: Double FlexAttention divides weights by `group_size` (mean vs sum over GQA groups). FlexAttention returns weights ~4x smaller than ground truth (group_size = n_head/n_kv_heads = 32/8 = 4). Triton score-sum matches PyTorch reference to ~1e-6 precision.                                                                                                                                                        

----


keys_values/flashinfer_wrapper.py
```diff
@@ -0,0 +1,1647 @@
...
+            # Each weights_chunk: (batch_size, n_query_groups, kv_len)
+            # Result: (batch_size, n_query_groups, kv_len)
+            weights = torch.stack(weights_chunks, dim=0).sum(dim=0)
+            
+            # Ensure weights are float32
+            weights = weights.float()
+        
+        return output, weights
+
+    def _fallback_sdpa(
```
mseeger, line 1271: I'd prefer not to have this, and deal with this choice in `attention.py`.

The best would be to limit the implementation here to what `flashinfer` can provide a fast implementation for, and do not support anything else.

keys_values/flashinfer_wrapper.py
```diff
@@ -0,0 +1,1647 @@
...
+            # Each weights_token: (batch_size, n_query_groups, kv_len)
+            # Result: (batch_size, n_query_groups, kv_len)
+            weights = torch.stack(weights_list, dim=0).sum(dim=0)
+            
+            # Ensure weights are float32
+            weights = weights.float()
+        
+        return output, weights
+
+    def _flashinfer_sdpa_long_sequence_chunking(
```
mseeger, line 1143: I am wondering whether we need this here?
The code outside deals with long sequences.

The kernel needs to support setups like `kv_len = 32k` and `q_len=4k` (say) with returning attention weights. The prefill case (so `kv_len == q_len`) is only needed if this is faster than the native PyTorch SDPA for this case. And I never need attention weights then.

keys_values/flashinfer_wrapper.py
```diff
@@ -0,0 +1,1647 @@
...
+    def scaled_dot_product_attention(
+        self,
+        query: torch.Tensor,
+        key: torch.Tensor,
+        value: torch.Tensor,
+        scale_factor: float,
+        return_attn_weights: bool = False,
+        token_positions: Optional[torch.Tensor] = None,
+        input_pos: int = 0,
+        sliding_window_size: Optional[int] = None,
```
mseeger, line 358: Please support this here only if the underlying flashinfer supports it. I really do not care about this, it is just something LitGPT supports. In the presence of sparse attention, this sliding window thing is not useful anyway.

keys_values/flashinfer_wrapper.py
```diff
@@ -0,0 +1,1647 @@
...
+        return q_len == 1 and kv_len > 1
+
+    def scaled_dot_product_attention(
+        self,
+        query: torch.Tensor,
+        key: torch.Tensor,
+        value: torch.Tensor,
+        scale_factor: float,
+        return_attn_weights: bool = False,
+        token_positions: Optional[torch.Tensor] = None,
```
mseeger, line 356: Please support this only if the underlying flashinfer supports it. Otherwise, we just use the reordering that I put in place for the others.

keys_values/flashinfer_wrapper.py
```diff
@@ -0,0 +1,1647 @@
...
+                token_positions, input_pos, sliding_window_size, chunk_size
+            )
+        except Exception as e:
+            logger.warning(f"FlashInfer SDPA failed with error: {e}. Falling back to eager implementation.")
+            return self._fallback_sdpa(
+                query, key, value, scale_factor, return_attn_weights,
+                token_positions, input_pos, sliding_window_size, chunk_size
+            )
+
+    def _flashinfer_sdpa(
```
mseeger, line 423: Main comment: Get rid of `_fallback_sdpa` and restrict arguments to only support cases for which the fast `flashinfer` applies.

scripts/benchmark_forward_only.py
```diff
@@ -0,0 +1,118 @@
+"""Forward-pass-only benchmark for H2O KV cache.
```
mseeger, line 1: Could go into `scripts`.

build_ext.py
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env python3
```
mseeger, line 1: Where is this being used?

setup.py
```diff
@@ -0,0 +1,96 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
mseeger: Does this work together with `pyproject.toml`? Did not have a `setup.py` so far

mseeger:
Please also run `./black.sh` and `./flake.sh` on your code to avoid `lint` errors.

test_compare_weights.py
```diff
@@ -0,0 +1,371 @@
...
+
+    # mask[b, h, q, k] = True if kv_pos <= query_pos
+    mask = tp_exp[:, :, None, :] <= q_positions[None, None, :, None]
+    scores.masked_fill_(~mask, float('-inf'))
+
+    # Softmax + sum
+    weights = torch.softmax(scores, dim=-1)  # [B, H, Q, KV]
+    # Sum over queries, then group heads -> [B, Hkv, KV]
+    weights = weights.reshape(batch, n_kv_heads, group_size, q_len, kv_len)
+    W = weights.sum(dim=(2, 3))
```
mseeger: Divide by `group_size` here.

To keep equal to the rest of the code, I define the reduction to be done with `mean`, not with `sum`.

This definition is to some sense arbitrary. It just means overall scores are divided by `group_size` (if avg is used instead of sum), and the order does not change.

The normal SDPA output does not depend on this definition.

Please also make sure to change the Triton code.

mseeger (changes requested):

keys_values/flashinfer_wrapper.py
```diff
@@ -0,0 +1,1647 @@
...
+                weights = tl.where(causal_ok, weights, 0.0)
+
+            w_acc += tl.sum(weights, axis=0)  # [BLOCK_KV]
+
+        # Write accumulated weights
+        w_ptrs = W_ptr + bh_id * W_stride_bh + kv_offsets
+        tl.store(w_ptrs, w_acc, mask=kv_mask)
+
+
+def triton_score_sum(
```
mseeger, line 154: Please divide by `group_size`, so reduce by avg, not by sum, to keep the same with remaining code.

mseeger (changes requested):

test_compare_weights.py
```diff
@@ -0,0 +1,371 @@
+"""Compare attention weights: double FlexAttention vs FlashInfer+Triton vs PyTorch reference.
```
mseeger: You can remove this test. I already included it now, and also fixed the issue for flex_attn.

Once your code is merged, I'll call it in the same test as well.

mseeger (changes requested):

test_compare_weights.py
vihangp: Removed.

keys_values/flashinfer_wrapper.py
vihangp: Division added in the flashinfer_wrapper.py.

vihangp:
> Please also run `./black.sh` and `./flake.sh` on your code to avoid `lint` errors.

This is done now.

build_ext.py
vihangp: This script works around CUDA version mismatch between system nvcc and the pytorch build by disabling the version check, and sets FORCE_CUDA_EXT flag. This makes sure that setup.py actually builds the extension.

mseeger:
Merging this, then fix issues and integration in subsequent PR.

PR #112, opened 2026-05-11
Add SM90 CUDA target for H100 support

Jantory:
This PR adds CUDA code generation support for NVIDIA H100 GPUs by including the `sm_90` architecture target in `build_ext.py`.

Previously, `_flashinfer_ops.so` was only compiled with kernels for A100 / `sm_80`. When running on H100 / `sm_90`, CUDA could not find a compatible kernel image at runtime, resulting in an error.

----

### DIFF
--- a/keys_values/finetune/args.py
+++ b/keys_values/finetune/args.py
@@ -171,9 +171,7 @@ class KVCacheArgs:
             New :class:`KVCacheArgs` object with `cache_kwargs` updated.
 
         """
-        new_cache_kwargs = {
-            name: getattr(self, name) for name in _CACHE_KWARGS_NAMES
-        }
+        new_cache_kwargs = {name: getattr(self, name) for name in _CACHE_KWARGS_NAMES}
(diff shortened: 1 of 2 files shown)


keys_values/finetune/args.py
```diff
@@ -171,9 +171,7 @@ def update_cache_kwargs(self) -> "KVCacheArgs":
             New :class:`KVCacheArgs` object with `cache_kwargs` updated.
 
         """
-        new_cache_kwargs = {
-            name: getattr(self, name) for name in _CACHE_KWARGS_NAMES
-        }
+        new_cache_kwargs = {name: getattr(self, name) for name in _CACHE_KWARGS_NAMES}
```
mseeger, line 174: If this comes from black. OK. Otherwise, please just run black.

mseeger (commented):
Vihang should give a comment, this is his code. I'll assign him.

vihangp:
Looks good to me.

PR #126, opened 2026-06-04
Add chunked log-prob computation for TRL GRPO integration

mmjerge:
### Motivation

This is the first step toward integrating KeysAndValues with HuggingFace TRL
for RL fine-tuning (GRPO/DPO) on long contexts. TRL's GRPOTrainer computes
per-token log-probabilities via a full forward pass, which OOMs on long
sequences. KeysAndValues can compute the same log-probs chunk by chunk through
a KV cache, with bounded GPU memory regardless of sequence length.

### What's included

- `keys_values/logprobs.py`: `chunked_per_token_logps()` — computes per-token
  log-probs using the existing chunked KV-cache forward infrastructure. Memory
  bounded by O(batch * chunk_size * vocab_size).
- `keys_values/finetune/grpo.py`: `GRPOLongContextTrainer` — subclass of TRL's
  `GRPOTrainer` that overrides `_get_per_token_logps_and_entropies` to route
  long sequences through the chunked path. Short sequences fall through to
  TRL's default (zero overhead).
- `test/test_logprobs.py`: 11 tests verifying correctness (chunked matches
  dense full-forward, temperature scaling, shapes, cleanup).
- `pyproject.toml`: Added `trl` as an optional dependency (`pip install keys_values[trl]`).

### Design decisions

- The log-prob utility (`logprobs.py`) has no TRL dependency — it's a general
  purpose function usable anywhere you need log-probs from long sequences.
- The trainer subclass lives in `finetune/` alongside the existing fine-tuning
  scripts.
- Threshold-based routing: only sequences exceeding `kv_cache_length` use the
  chunked path.

### Next steps

Full end-to-end pipeline (generation + model adapter bridging keys_values GPT
with HuggingFace's PreTrainedModel interface) will follow in a subsequent PR.

---

### DIFF
--- a/keys_values/finetune/grpo.py
+++ b/keys_values/finetune/grpo.py
@@ -32,6 +32,7 @@ Usage::
     )
     trainer.train()
 """
+
 from __future__ import annotations
 
 import torch
@@ -127,6 +128,7 @@ class GRPOLongContextTrainer(GRPOTrainer):
             else None
         )
         return logps, entropies
+
 
 def _unwrap(model) -> GPT:
     """Peel wrappers (DDP, PEFT, HF, ...) until we hit a ``GPT`` instance."""
--- a/keys_values/finetune/grpo.py
+++ b/keys_values/finetune/grpo.py
@@ -1,0 +1,152 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License").
+# You may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+GRPO fine-tuning with KeysAndValues long-context KV cache support.
+
+Provides ``GRPOLongContextTrainer`` — a subclass of TRL's ``GRPOTrainer``
+whose per-token log-probability computation uses KeysAndValues' chunked
+KV-cache forward pass, bounding GPU memory for arbitrarily long sequences.
+
+Usage::
+
+    from keys_values.finetune.grpo import GRPOLongContextTrainer
+
+    trainer = GRPOLongContextTrainer(
+        model="Qwen/Qwen2.5-0.5B-Instruct",
+        reward_funcs=my_reward_func,
+        train_dataset=dataset,
+        kv_cache_name="h2o-torch-quantized8",
+        kv_cache_length=16384,
+        kv_chunk_size=1024,
+    )
+    trainer.train()
+"""
+from __future__ import annotations
+
+import torch
+from trl.trainer.grpo_trainer import GRPOTrainer
+
+from keys_values.logprobs import chunked_per_token_logps
+from keys_values.model import GPT
+
+_UNWRAP_ATTRS = ("gpt_model", "model", "base_model", "module")
+
+
+class GRPOLongContextTrainer(GRPOTrainer):
+    """``GRPOTrainer`` with KV-cache chunked log-prob computation.
+
+    Overrides TRL's full-sequence forward with KeysAndValues' bounded-memory
+    chunked path for sequences exceeding ``kv_cache_length``. Short sequences
+    fall through to TRL's default — zero overhead.
+
+    Parameters
+    ----------
+    kv_cache_name : str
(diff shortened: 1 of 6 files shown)


mseeger (commented):
Hello, thanks a lot for this PR. If this enables GRPO to work, this is great. With these changes, is GRPO already working, or are extra steps needed?

In general, I am wondering about one thing. Given that `LongContextInferenceModel` supports chunked forward computations already, would it not be possible to get what you do here by coding a specific "loss function", which instead of a loss value computes log probs and writes them out (as chunks)? This may even be doable by wrapping `keys_values.head_model.CrossEntropyOnLogits`.

Just wonder what you think. Such a solution would need very little new code, and you'd not have to duplicate code which is in `LongContextInferenceModel` already.

mmjerge:
Thanks for the thoughtful comments. 

Regarding Question 1: 

GRPO is not end-to-end yet. This PR covers step 1 of 3.

1. Log-prob computation (this PR) — bounded-memory per-token log-probs via chunked forward
2. Generation — TRL calls model.generate() expecting a HuggingFace interface. keys_values.generate.batched_generate_fn already does long-context generation through the KV cache, but it's not wired to TRL's calling convention yet. This is a thin adapter (next PR).
3. Backward through GRPO loss — TRL calls loss.backward() on the policy gradient loss. For long sequences this needs LongContextGradientModel to keep memory bounded during backprop. This is the most involved piece.

I was trying to break up these parts into different MRs to make it more manageable/modular, but if you would prefer having the full pipeline in a single MR, I can do that. 

Regarding Question 2: 

Your suggestion is a much better design pattern. I've refactored to do this. The new implementation is a LogProbsHeadModel (same interface as CrossEntropyOnLogits) that gathers per-token log-probs instead of reducing to a scalar loss. compute_logprobs() just plugs it into LongContextInferenceModel.forward(). Good idea!

keys_values/finetune/grpo.py
```diff
@@ -0,0 +1,152 @@
...
+                cache_length=self.kv_cache_length,
+                chunk_size=self.kv_chunk_size,
+                cache_kwargs=self.kv_cache_kwargs,
+                temperature=self.temperature,
+                compute_entropy=compute_entropy,
+            )
+            for i in range(0, input_ids.size(0), bs)
+        ]
+
+        logps = torch.cat([r[0] for r in results])
```
mseeger, line 124: What is a bit surprising: This combined tensor will also end up in GPU memory. Will this fit for long contexts?
I wonder if it is possible to somehow yield the individual chunks back to GRPO, so that the full combined tensor is never built, at least not in GPU memory.

But we can leave this for later.

mseeger (dismissed):

mseeger:
@mmjerge , please run `./black.sh` and `./flake.sh`, so that the linter tests pass.

mmjerge:
Ran ./black.sh and ./flake.sh and tests passed. 

(keyval_venv) mjerge@80a99709b43d keys_values % source key
val_venv/bin/activate && ./black.sh 2>&1 | tail -5
Warning: Python 3.12 cannot parse code formatted for Python 3.15. To fix this: run Black with Python 3.15, set --target-version to py312, or use --fast to skip the safety check. Black's safety check verifies equivalence by parsing the AST, which fails when the running Python is older than the target version.
All done! ✨ 🍰 ✨
158 files left unchanged.
(keyval_venv) (base) mjerge@80a99709b43d keys_values % sou
rce keyval_venv/bin/activate && ./flake.sh 2>&1 | tail -5
<unknown>:164: SyntaxWarning: invalid escape sequence '\s'
<unknown>:205: SyntaxWarning: invalid escape sequence '\s'
./keyval_venv/lib/python3.12/site-packages/sympy/polys/numberfields/resolvent_lookup.py: "pyflakes[F]" failed during execution due to RecursionError('maximum recursion depth exceeded')
Run flake8 with greater verbosity to see more details
0

keys_values/finetune/grpo.py
mmjerge: Good catch. The thing that, I think, makes this okay is that the combined tensor is the per-token log-probs. So, its shape is (batch_size, completion_length), i.e. one scalar per token, not per vocab entry, which my back of the envelope math shows its only about 1 MB. The tensor that would actually blow up memory is the (batch, chunk, vocab_size) logits, and that one already gets chunked away inside compute_logprobs, so it never fully lands in GPU memory. So I think the torch.cat is fine here even for long contexts. 

You're right that there's a nicer streaming version, though. The catch is that TRL's _get_per_token_logps_and_entropies wants a single tensor back, so yielding chunks would mean changing something on the TRL side or going through a custom loss path, which we can do but more effort. I think it probably makes sense to revisit that when we get to the full pipeline.

mseeger (approved):
Thanks for the PR, will merge this now. Looking forward to the other 2, and making GRPO work.

PR #128, opened 2026-06-15
End-to-end GRPO with KeysAndValues KV cache

mmjerge:
## Phase 2: End-to-end GRPO with KeysAndValues

This builds on the merged log-prob work (#126) to complete the GRPO loop, so
that generation, scoring, and the policy gradient all run through the KV cache.
GPU memory stays bounded even for long completions.

### What is included

- `keys_values/generate/trl_rollout.py`: `generate_completions`, completion
  generation via chunked KV-cache decode (wraps the existing `batched_generate_fn`).
- `keys_values/finetune/grpo_loss.py`: `GRPOLossHeadModel`, the GRPO clipped
  surrogate loss expressed as a `HeadModel`. This lets the policy gradient flow
  through `LongContextGradientModel` (memory-bounded backward), exactly the way
  `CrossEntropyOnLogits` does for supervised fine-tuning.
- `keys_values/finetune/grpo_loop.py`: `grpo_step` and `compute_group_advantages`,
  a standalone end-to-end GRPO loop on a `keys_values.model.GPT`.
- `keys_values/logprobs.py`: added a backward-compatible `verbose` parameter.
- `examples/trl_grpo_demo.ipynb`: runnable CPU demo of the full pipeline.
- Tests: end-to-end `grpo_step` and the generation adapter (19 passing on CPU).

### Design note

The GRPO loss is a `HeadModel`, so the gradient reuses the existing
`LongContextGradientModel` machinery rather than duplicating any forward/backward
logic. The `grpo_step` loop runs directly on a keys_values `GPT` rather than
through TRL`s `GRPOTrainer` (which expects a HuggingFace `PreTrainedModel`). Happy
to add a `PreTrainedModel` adapter as a follow-up if you would prefer driving this
from `GRPOTrainer` directly.

### AI usage

AI was used to generate docstrings, documented in `ai_dev/trl_integration.md`.

---


keys_values/rl/grpo/loop.py
```diff
@@ -0,0 +1,216 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
mseeger, line 1: Could you add some docs (markdown file), in `docs`), explaining what needs to be installed, and how this can be run?

mseeger: For example, is KeysAndValues used (1) for inference, (2) for gradient updates, or (3) for both?

mseeger (commented):
I think it would be best to move all these files into a module `keys_values/rl`, maybe even `keys_values/rl/grpo`.

This would also immediately show any user that we support RL.

keys_values/rl/grpo/loop.py
mmjerge: Added [`docs/grpo_integration.md`](../blob/grpo-phase2-generation/docs/grpo_integration.md) in commit 0c38abd. It covers installation (base install plus `pip install -e .[trl]` for the TRL trainer path) and how to run both entry points — the standalone `grpo_step` loop and the `GRPOLongContextTrainer`, with a pointer to the example notebook.

mmjerge: Good question — the docs now spell this out explicitly. Short version:

- **Standalone loop (`grpo_step`)**: the KV cache is used for **both**. It backs generation and the old-log-prob scoring (inference), *and* the policy-gradient pass via `LongContextGradientModel` + `GRPOLossHeadModel` (memory-bounded backward).
- **`GRPOLongContextTrainer` (TRL path)**: the cache is used only for the per-token log-prob computation TRL depends on (reference and policy), and only for sequences longer than `kv_cache_length`; TRL still owns generation and the optimizer step.

There is a per-stage table in `docs/grpo_integration.md` under "Where the KV cache is used".

mmjerge:
Thanks for the review @mseeger — addressed all three points in commit 0c38abd:

**1. Move into a dedicated RL module.** Agreed, this makes RL support much more discoverable. The GRPO code now lives under `keys_values/rl/grpo`:

```
keys_values/rl/
  logprobs.py            # shared memory-efficient per-token log-probs
  grpo/
    trainer.py           # GRPOLongContextTrainer (was finetune/grpo.py)
    loop.py              # grpo_step standalone loop (was finetune/grpo_loop.py)
    loss.py              # GRPOLossHeadModel (was finetune/grpo_loss.py)
    rollout.py           # generate_completions (was generate/trl_rollout.py)
```

All imports, tests (now under `test/rl/`), the example notebook, and the AI-dev notes were updated. Moves were done with `git mv` so history is preserved.

**2. Docs.** Added `docs/grpo_integration.md` covering what to install and how to run both entry points (standalone loop and the TRL trainer).

**3. Where the KV cache is used.** Documented per-stage — replied inline as well. In short: the standalone loop uses it for both inference (generation + old-log-prob scoring) and the gradient pass; the TRL trainer uses it only for the long-sequence log-prob computation.

`test/rl/` passes (19 tests, CPU). Let me know if you would prefer different file names within `rl/grpo` or a flatter layout.

mseeger:
Thanks, great. I'll merge this.

PR #133, opened 2026-06-18
Add vLLM (V1) KV cache policy integration: lastrec bridge + spike

mmjerge:
## Summary
First slice of integrating keys_values selective KV cache policies into
vLLM 0.23 (V1 engine). Draft / work-in-progress: lands the foundation and the
first policy (lastrec).

Included:
- `docs/vllm_integration.md`: design notes, V1 extension-point map, and the H2O
  architecture decision (block-level approximation vs. dense-cache backend).
- `examples/vllm_spike.py`: feasibility spike (runs a baseline model on vLLM V1,
  dumps the resolved attention backend + KV cache geometry).
- `keys_values/vllm/`: bridge package — policy config, registration via
  `KVCacheSpecRegistry`, and the lastrec `KVCacheSpec` + manager (built on
  vLLM's sliding-window machinery).
- `test/vllm/`: no-engine unit tests against vLLM's pure-Python `BlockPool`,
  plus registration smoke tests.

## Tested
- `test/vllm/` passes on an A10G (vLLM 0.23 / torch 2.11): lastrec spec/manager,
  bounded-footprint and sliding-window-parity properties, registry routing.
- Spike runs Qwen2.5-0.5B end to end on V1.

## Not yet included (follow-ups)
- Serve-time wiring so a model actually uses `LastRecSpec`.
- H2O (attention-weight eviction) — design documented, implementation pending.
- Parity vs. the LitGPT path; long-context eval.

## AI usage
See `ai_dev/vllm_integration.md`.


docs/vllm_integration.md
```diff
@@ -0,0 +1,170 @@
+# vLLM Integration
+
+Status: **early spike** (branch `vllm-integration`)
+
+Target vLLM version: **0.23.0** (latest), running on the **V1 engine**.
+
+> History: this was originally scoped against vLLM 0.6.5 (the version in the
```
mseeger, line 7: This sounds reasonable to me, let us integrate to the most recent (stable) version of vLLM, and not try to be compatible with older versions.

What is the drawback here? Is it harder to integrate against the more recent oine?

docs/vllm_integration.md
```diff
@@ -0,0 +1,170 @@
...
+
+## Goal
+
+Bring this library's *selective KV cache policies* (H2O, qH2O, smart-lastrec,
+lastrec, ...) to vLLM, so users can run long-context inference with advanced
+cache eviction while benefiting from vLLM's serving stack. Reuse vLLM's fast
+kernels and multi-device machinery; plug in our cache abstractions and eviction
+policies.
+
+This is **not** primarily a "register a new model" task. vLLM ships its own
```
mseeger, line 47: I like this. Let is force as little "non-vLLM" stuff on users. They should get the good (and say unique) part of what is provided here, but ideally nothing that vLLM already has and has better.

docs/vllm_integration.md
```diff
@@ -0,0 +1,170 @@
...
+
+Verify: `python -c "import torch, vllm, litgpt, keys_values.kvcache.base"`.
+LitGPT (`torch>=2.7`) and vLLM 0.23 (torch 2.11) coexist; the keys_values
+`kvcache/` module is pure torch (zero litgpt imports), so it slots into either.
+
+## The impedance mismatch (still the core problem)
+
+| Concern | keys_values | vLLM V1 |
+|---|---|---|
+| Cache layout | dense `(batch, n_query_groups, cache_length, head_size)` + `token_pos` | paged blocks; per-layer `KVCacheSpec` (block_size, num_kv_heads, head_size, dtype) |
```
mseeger, line 79: I only have a rough understanding of PagedAttention memory layout. In my mind:
- I am using the dense form, so `(batch, n_query_groups, cache_length, head_size)`. This can be different for every layer. For example, `cache_length` can be different. Not sure about `n_query_groups`. I am also using `token_pos` of shape `(batch, n_query_groups, cache_length)` to state what is where.
- They probably use memory chunks of shape `(length, head_size)`, indexed by batch and head dim, and these do not have to be part of one dense array.

One potential advantage of their setup is they could have different `cache_length` per batch and head dimension. I cannot have that.

docs/vllm_integration.md
```diff
@@ -0,0 +1,170 @@
...
+| Eviction granularity | per `(batch, head, slot)` | per block, via a `KVCacheManager` / block pool |
```
mseeger, line 80: OK, but what is a "block"? What indexes does it have?

docs/vllm_integration.md
```diff
@@ -0,0 +1,170 @@
...
+| Eviction decision | inside `KVCache.forward`, H2O needs **attention weights** | block managers decide by position/recency; kernels don't return weights |
```
mseeger, line 81: For now, I recommend to start with logics which do not need these weights. The most advanced one would be the `smart-lastrec`, which I currently refactor to better work with left padding.

docs/vllm_integration.md
```diff
@@ -0,0 +1,170 @@
...
+| Attention call | `KVCache.forward(query, key, value, token_idx)` does evict + SDPA | model calls `Attention` layer; V1 attention backend + metadata builder |
+
+Two hard constraints persist:
+1. **Attention weights.** H2O scores slots by summed attention weight; vLLM
+   kernels don't expose them. keys_values already solves this for its own stack
+   (`keys_values/attention/`, `kvcache/attn_weights.py` — vendored FlashInfer +
+   Triton score-sum). That has to be reachable from the vLLM attention path.
+2. **Granularity.** vLLM evicts whole blocks; our policies score per `(head,
```
mseeger, line 89: Need to figure out the granularity of blocks.

mseeger:
Please also look at the failed tests. Also make sure to run black

keys_values/vllm/managers.py
```diff
@@ -0,0 +1,43 @@
...
+rule explicitly. A distinct class lets the registry route ``LastRecSpec`` here
+and gives smart-lastrec a place to extend.
+"""
+
+from __future__ import annotations
+
+from vllm.v1.core.single_type_kv_cache_manager import SlidingWindowManager
+
+
+class LastRecManager(SlidingWindowManager):
```
mseeger, line 40: I think the next step could be integrating the `smart-lastrec` KV cache logic. This is really something that is not in vLLM already, and it makes decisions with depend on batch dimension (but not on head position).

But it does not need summed attention weights, so we get around that one for now.

mmjerge: I agree with this

docs/vllm_integration.md
mmjerge: Makes sense, I will change that.

docs/vllm_integration.md
mmjerge: Added a "Why target the latest stable, and what it costs" section in a2587a3. Short version: integrating against the latest is *easier* on the axis that matters (deps — a recent vLLM resolves onto a recent torch, so one env holds vLLM + LitGPT + keys_values). The real cost isn't difficulty, it's that the V1 seams we hook (KVCacheSpec, the manager family, get_kv_cache_spec, the platform hook) are engine internals that shift between releases. So we pin 0.23.0, bump deliberately, keep the bridge surface small, and guard with parity tests.

docs/vllm_integration.md
mmjerge: Captured this as an explicit "Design principle — minimize what we impose on users" note in a2587a3: reuse vLLM for paging/scheduling/kernels/model loading, ship only the part that's genuinely ours (the selective eviction policies), keep the bridge torch-only and litgpt-free, and aim for a single policy/config flag as the user-facing surface.

docs/vllm_integration.md
mmjerge: Documented this in the new "vLLM paged KV layout" section (a2587a3). The key point for the mental-model comparison: per-request length *can* vary (each sequence owns its own block table and block count), but per-head length *cannot* — block_size and the block table are shared across all KV heads in a layer. So vLLM is actually more flexible than the dense form on the batch axis, but strictly less flexible on the head axis, which is exactly why faithful per-head H2O isn't possible at block granularity.

docs/vllm_integration.md
mmjerge: Added a precise definition in a2587a3. A block is a slab holding block_size consecutive tokens' K and V for all KV heads of one layer — physical shape per layer (2, num_blocks, block_size, num_kv_heads, head_size) (16/2/64 in the spike). To address a single KV vector you need (layer/group, physical block_id, offset in [0, block_size), kv_head, feature). A request owns a block table per group (an ordered list of physical block_ids); a logical token t maps via physical_block = block_table[t // block_size], offset = t % block_size.

docs/vllm_integration.md
mmjerge: Answered in the "Granularity" part of the new section (a2587a3): the smallest allocatable/evictable unit is one block = block_size tokens (16 in the spike) for a whole layer across all KV heads, uniform within a KV cache group. It's a per-model config value, not something we choose per policy.

docs/vllm_integration.md
mmjerge: Agreed — reordered the roadmap in a2587a3 so smart-lastrec is the explicit next step before H2O: positional, no attention weights, and batch-dependent (per-request initial/grace regions) but not head-dependent, which fits the paged block model. Noted that upstream smart-lastrec is being refactored for left padding and we'll align the bridge once that lands.

keys_values/vllm/managers.py
mmjerge: Added a note in managers.py and the doc roadmap (a2587a3): SmartLastRecManager is the planned next subclass of LastRecManager — weight-free, batch- not head-dependent — so it slots into the paged model without per-head divergence. Will pick it up after the upstream left-padding refactor.

docs/vllm_integration.md
mseeger: This is a bit annoying. However, `smart-lastrec` does not need this flexibility on head axis.

I'll read their paper to understand this.

docs/vllm_integration.md
mseeger: And you are sure the block needs to stretch over all heads? It cannot be a subset?

docs/vllm_integration.md
mseeger: Footnote in vLLM paper:
"In Transformer, each token has a set of key and value vectors across layers
and attention heads within a layer. All the key and value vectors can be
managed together within a single KV block, or the key and value vectors at
different heads and layers can each have a separate block and be managed
in separate block tables. The two designs have no performance difference
and we choose the second one for easy implementation."

At least there, they say that different heads can have different blocks.

mseeger:
Hi Michael, I am reading the vLLM paper. At least from the paper:
- Block contains KV information (key and value vecs) for a fixed number of tokens and a single head and layer. At least, it says nothing about a block extending across all heads
- On the other hand, I find nothing about removing a block or overwriting it

In the standard inference case, KV cache always grows until the end, then all blocks for one job are freed.

In sparse attention, we'd overwrite stuff. With H2O, the entries to be overwritten could be in all sorts of blocks. That would be a problem. One would have to score whole blocks. Now, if blocks are not too large, this could still make sense, right? Given the H2O scores per (b, h, j), we could map them to blocks and remove/overwrite those which score lowest.

Could you check again about the thing with the head? The paper explicitly says different heads can be in different blocks.

mseeger:
The other issue, as you say, is of course the missing summed attention weights. But this is at least something we could ask them as a feature. We cannot ask them to change their memory layout, that is the whole point of vLLM!

docs/vllm_integration.md
mmjerge: Good catch — you're right, and I've corrected the doc in 2f63904. The earlier wording ('a block must span all heads') overclaimed. The paper is explicit that both layouts are valid and interchangeable, and that they chose the separate-block-per-head-and-layer design for ease of implementation, so per-head block tables are not ruled out in principle.

The constraint I actually depend on is narrower and empirical: vLLM's **shipped V1 attention specs** (FullAttentionSpec, SlidingWindowSpec, ...) use the all-heads-in-one-block layout with a single block table per KV-cache group. The spike confirms this — the per-layer KV tensor is (2, num_blocks, block_size, num_kv_heads, head_size) with heads as an axis *inside* the block, and one block table per request. So for the layout we integrate against, a block is the smallest unit across all heads, even though PagedAttention as a design permits per-head blocks.

Getting true per-head eviction would mean defining per-head KV-cache groups/specs, which is a much larger change to vLLM's manager + metadata machinery and out of scope for this bridge. The doc now states this distinction explicitly instead of presenting it as a hard law. And per your other note, smart-lastrec doesn't need head-axis flexibility anyway, so the near-term path is unaffected.

mmjerge: Thanks — agreed, and I've folded this into the doc (9217c57): cited Kwon et al. (SOSP 2023) and corrected the framing so the paper's core invariant is just *fixed-size blocks* (per-(b,h) blocks holding several tokens are permitted), with the all-heads-in-one-block layout called out as vLLM's *shipped* V1 choice rather than a law.

On the path forward, I've recorded the decision explicitly: for vLLM we do H2O at **block granularity** — aggregate scores over a block's tokens (and over heads in the group, since the shipped specs share one block table) and evict/overwrite whole blocks — instead of single-token KV eviction. No per-head/per-token fidelity inside vLLM; the dense keys_values stack stays the reference for that, and Option B (dense backend) is demoted to a fallback. The one genuinely hard piece left is mechanical: vLLM's managers only free a contiguous prefix, while H2O frees arbitrary middle blocks (Blocker 2 in the doc). Phase order is unchanged — lastrec (done) → smart-lastrec (weight-free, next) → block-level H2O.

mmjerge:
On the summed attention weights: good news — I don't think we need to ask vLLM for a feature, and we definitely don't have to touch their memory layout. vLLM's **V1 FlashInfer backend** already runs its wrappers with `return_lse=True` (`vllm/v1/attention/backends/flashinfer.py`), so it hands back the per-query log-sum-exp. With LSE + Q + K we can reconstruct per-position attention mass as `exp(q_i·k_j·scale - lse_i)` and sum it — the same recipe the keys_values FlashInfer path already uses, summable with the existing Triton score-sum kernel. So the plan is: force the FlashInfer backend, capture LSE per layer, compute the per-position score-sum, and aggregate to block scores for eviction. This is written up as 'Blocker 1 — attention weights in-engine' in docs/vllm_integration.md, and step 1 (FlashInfer selectable via the 0.23 `attention_backend` knob, hooks fire, output stays correct) is confirmed on a g5.xlarge.

The genuinely hard part isn't the weights — it's Blocker 2: vLLM's built-in managers only ever free a contiguous *prefix*, while H2O frees arbitrary *middle* blocks, which the append-only block table doesn't natively represent. That's where the remaining work is, and it stays entirely on our side of the bridge (no memory-layout change asked of vLLM).

mmjerge:
@mseeger this is now marked ready for review. I've addressed all the outstanding threads:
- Corrected the per-head/block claim and cited Kwon et al. (SOSP 2023); recorded the agreed block-level H2O direction in docs/vllm_integration.md.
- Summed attention weights: covered via the FlashInfer LSE path (Blocker 1) — no vLLM feature request or memory-layout change needed.
- black is clean and the CI test failures from the earlier pass are resolved.

Could you take another look when you have a chance? (I don't have permission to formally request your review on this repo, hence the ping.)

mseeger:
Hi Michael. First, if you see a way for how to get the summed attention weights in current vLLM, that is very exciting, and I'd encourage you to give it a try. I can have a look at code once you have it.

I have a way of getting them with 2 FlexAttention calls, that is implemented in the library. But that needs 2 calls of FlexAttention, not just 1.

keys_values/vllm/managers.py
mseeger: The upstream `slr` refactor is done. It does not buy much, rhough.

mseeger:
To me, the fact that a block must extend over all heads, seems like a grave design mistake. Are you really sure?
Maybe this has to do with how they do the SDPA inside.

Now, if that is a restriction of PagedAttention, I do not see how one would do H2O in there. Sure, we could evict blocks based on the H2O score, but this is normally done per head, even in the original H2O paper.

mseeger:
https://discuss.vllm.ai/t/sparse-attention-e-g-h2o/2745

You are right. This is indeed something that does not work, and that has been asked before.

mmjerge:
Thanks for digging into this and for the forum thread — that settles the block/head question, and I've folded your findings into the doc (f14c003).

**Per-head eviction / block spanning all heads.** Confirmed and now cited: I added an "Upstream confirmation" note that references your forum thread and the three RFCs it surfaced ([#10646](https://github.com/vllm-project/vllm/issues/10646), [#12254](https://github.com/vllm-project/vllm/issues/12254), [#5751](https://github.com/vllm-project/vllm/issues/5751)). To your "is this a grave design mistake / to do with SDPA?" — it's an implementation choice tied to how the paged kernels index KV (one shared block table per layer/group, heads as an inner axis), and per the RFCs, true per-head sparse KV would need a block-manager + memory-layout redesign. So block-level H2O is the right target for current vLLM; genuine per-head eviction waits on one of those RFCs landing. The dense keys_values stack stays the per-head reference in the meantime.

**Summed attention weights — we already have code, and it's a single call.** This is in the PR at `keys_values/vllm/attention.py`: `reference_summed_attention` (pure-torch contract, unit-tested in `test/vllm/test_attn_weights.py`) and `summed_attention_weights`, which routes through keys_values' FlashInfer + Triton score-sum path and returns `(batch, n_kv_heads, kv_len)` summed weights. The nice part re: your point about needing 2 FlexAttention calls — this gets the weights in a **single** FlashInfer SDPA call (`return_attn_weights=True`), because vLLM's V1 FlashInfer backend already runs with `return_lse=True`, so the LSE needed to normalize is on the same call. No feature request to vLLM, no second pass. There's also an in-engine prototype on the follow-up branch (`examples/vllm_h2o_score_probe.py`) that gathers a request's paged K/V and computes correct per-position mass during decode. Happy to walk you through that code or open it as a separate PR for review — whichever is easier for you.

The genuinely remaining work is the wiring (mapping per-position scores to blocks and handling arbitrary middle-block eviction against vLLM's append-only block table), not obtaining the weights.

keys_values/vllm/managers.py
mmjerge: Good — thanks for confirming the slr refactor landed. I've updated the doc note here so it no longer lists an upstream dependency: smart-lastrec is the next policy to integrate (weight-free, batch- not head-dependent, so it fits the paged block model), and I'll implement SmartLastRecManager as the LastRecManager subclass this scaffold was built for. Noted that the refactor 'does not buy much' — I'll align to the current upstream shape rather than wait on anything.

mmjerge:
@mseeger — as offered, I split the in-engine H2O attention-weight work into its own PR so you can review that code directly, separate from this one:

**https://github.com/mmjerge/keys_values/pull/2**

It's stacked on this branch (`vllm-integration`), so the diff there is just the incremental H2O-wiring code:
- FlashInfer backend selection + layer hooks (`examples/vllm_h2o_probe.py`)
- `FlashInferImpl.forward` surface mapping (`examples/vllm_h2o_lse_probe.py`)
- **in-engine per-position score computation** for single-seq decode (`examples/vllm_h2o_score_probe.py`) — gathers the request's paged K/V and computes correct per-KV-position attention mass using the task-3 `reference_summed_attention` contract
- a `docs/vllm_integration.md` "H2O wiring (task 4.3)" section documenting Blocker 1 (weights in-engine — tractable, single FlashInfer call) and Blocker 2 (arbitrary middle-block eviction — the hard part)

The weight math itself (single-SDPA-call extraction + tests) is already in *this* PR under `keys_values/vllm/attention.py`. Happy to walk through any of it.

keys_values/vllm/managers.py
```diff
@@ -0,0 +1,141 @@
...
+        return max(0, num_computed_tokens - self.sliding_window + 1)
+
+
+from vllm.utils.math_utils import cdiv
+from vllm.v1.core.single_type_kv_cache_manager import FullAttentionManager
+
+from keys_values.vllm.scores import BlockScoreTracker, select_evictions
+
+
+class H2OManager(FullAttentionManager):
```
mseeger, line 59: If I understand this correctly, it circumvents the vLLM internals entirely, by using SDPA as implemented in KeysAndValues.

mseeger (commented):
For this to continue, we need to be sure about the goal. This could be:

1. Plug KeysAndValues attention with KV caching into vLLM as the outer inference engine
2. Plug KeysAndValues KV cache logics into vLLM, but use all underlying vLLM primitives for SDPA and such

My preference is for 2, because vLLM brings so much low-level stuff to the table. For 1, I'd have to be convinced that if attention is fully run through KeysAndValues, there is still an advantage of running vLLM.

There WOULD be (for 1) if H2O could be run properly in vLLM, because that is missing there right now.

But their memory management seems to preclude sparse attention as H2O needs. This is something I did not know when suggesting this direction.

In order to justify spending more effort on vLLM integration, let us try to be specific about the benefits:
- What would work with the integration that would not work without it?
- Which KV cache policies from KeysAndValues can we run?

Otherwise, let us focus more on getting RL to work and explore the bottlenecks of RL with sparse attention!

mseeger:
@mmjerge , can I close this one?

We have now learned that sparse attention of the H2O and related types cannot be integrated into vLLM, due to their paged memory layout, and especially the constraint that each page needs to run over all heads.

Sorry you spent quite some work on this one.

mmjerge:
Agreed, closing. The conclusion is sound: H2O-style per-head eviction is fundamentally incompatible with vLLM's paged KV layout, where each page must span all heads — the retained-token sets diverge per head, so there's no consistent page granularity. No wasted effort on my end; ruling this out is what pointed the work toward the RL-at-long-context direction, which is where the results are now (#142). Thanks for the forum digging that settled it.

Issue #96, opened 2026-04-29
Test comparing attention weights fails for FlashInfer if `q_len=1`


mseeger:
**Describe the bug**  

Look at `test/attention/test_attn_weights.py`, `test_larger_comparison`, the `q_len=1` cases.

Currently, FlashInfer is skipped for this one. If the skip is removed (in `get_variants`), the tests fail for FlashInfer. All other variants give the same results.

FlashInfer works for all other `q_len` values. Likely this is because the case `q_len=1` is implemented differently.

mseeger:
Note: Please wait for #95 to be merged!

mseeger:
OK, this IS a problem: For now, our evaluation scripts refuse to use FlashInfer SDPA, because the `q_len=1` case is needed for token generation, and generation fails right now (giving very poor results).

Training is not affected, since chunks are never just size 1.

mseeger:
Resolved by #136

PR #136, opened 2026-07-07
Fix FlashInfer attention weights for q_len=1 (#96)

vihangp:
The q_len=1 decode path with return_attn_weights=True routes to the tiled CUDA kernels, which default the query position to 0 when input_pos is not given. With causal=True this masked out every KV entry except position 0, producing one-hot attention weights and breaking token generation with H2O-style cache strategies.

Since the single query token is the last position, causal masking is a no-op (it must attend to the whole KV cache). Pass causal=False when return_weights=True. The return_weights=False path keeps causal=True so it still dispatches to the fast native FlashInfer decode kernel.

Also remove the q_len > 1 exclusion for FlashInfer in test_attn_weights.py; all q_len=1 cases now pass.

> REPLACE THIS TEXT BLOCK
>
> Describe the reason for this change, what the solution is, and any
> important design decisions you made. 
>
> Remember to follow the [CONTRIBUTING GUIDE] for any code you submit.
>
> [CONTRIBUTING GUIDE]: https://github.com/awslabs/keys_values/blob/main/CONTRIBUTING.md

Closes #<issue number here>.

----

### DIFF
--- a/keys_values/attention/flashinfer_ops.py
+++ b/keys_values/attention/flashinfer_ops.py
@@ -247,6 +247,14 @@ def sdpa_decode(
             "Vendored FlashInfer kernels are not available. " f"Error: {_load_error}"
         )
 
+    # The single query token is the last position, so causal masking is a
+    # no-op: it must attend to all of the KV cache. With return_weights=True,
+    # the tiled kernels are used, which default the query position to 0 if
+    # `input_pos` is not given, so `causal=True` would mask out everything
+    # except KV position 0 (issue #96). Pass `causal=False` in this case.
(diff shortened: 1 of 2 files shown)


mseeger:
Could you rebase to `main`? I pushed something that was not correctly formatted.

Can you check whether the test works now that was commented out?

mseeger:
I put in some code somewhere which ensures that FlashInfer is not used if `q_len == 1`. But now, I cannot find it. Maybe you see where this is?

mseeger:
Nevermind, maybe I forgot to do this. Will have to rerun all evals using H2O now.

Issue #62, opened 2026-03-14
Backward CPU offloading: Asynchronous transfer


mseeger:
**Is your feature request related to a problem? Please describe.**
CPU offloading in `backward` works by iterating this over layers:
- Load weights from CPU
- Run backward from head gradients
- Store gradients to CPU

These steps are currently run sequentially.

**Describe the solution you'd like**
Can we run these steps in parallel?

- Switch between two shards on GPU
- Run backward on one while loading weights for the other, and storing gradients for the previous

Needs clear understanding how async CPU <-> GPU transfer works! We know how transfer between GPUs works.

mseeger:
The issue title is misleading: CPU-GPU transfer plays a key role in both `forward` and `backward`. Here are all possible cases:

* `forward`: Write layer input checkpoints to CPU
* `forward`, KV cache offloading: CPU-GPU bidirectional streaming
* `backward`: Load layer input checkpoints from CPU
* `backward`, head gradients: CPU-GPU bidirectional streaming
* `backward`, weights/gradients offloading: CPU-GPU bidirectional streaming

There are two steps, do one after the other:

1. Decouple CPU-GPU transfer from GPU compute, using 3 streams: load (CPU->GPU), compute, store (GPU->CPU)
2. Ensure (by cycling buffers) that 3 streams do not interfere (load target is read in compute; store source is written to in compute)

Step 2 is harder in `backward`, where we have less free GPU memory.

mseeger:
Here is a plan. For the naming:

* layer is clear
* chunk: Sequence batch split into chunks. Unit passed through the model, unit of KV cache updates
* cell: List of neighboring chunks, until total size about cache length. Forward passes loop over cells (outermost), then layers, then chunks of cell (innermost). Activations for one cell fit in GPU memory, but not activations for complete sequence.

### Implements 3 streams solution: load (CPU->GPU), compute (GPU), store (GPU->CPU)

Follow code examples posted above. Don't use Transformer Engine dependence.

Implement tests that this works if there is no interference:
* load target is read by compute
* store source is written by compute
* load target overlaps with store source

The following should be OK:
* store source is read by compute (both are read accesses)

### `forward`: Write layer input checkpoints to CPU

Cell by cell. store and compute work on the same layer input. Synchronization at the end of each cell. No interference.

### `backward`: Load layer input checkpoints from CPU

- Use ring of 2 arrays
- In parallel: load for next cell needed; compute for current cell

[...]

PR #138, opened 2026-07-10
Async cpu transfer

vihangp:
Adds opt-in async CPU↔GPU transfer during gradient computation (#62), behind
`--grad.async_cpu_transfer` (default `False`; requires pinned checkpoint
memory). Layer-input checkpoints, KV-cache checkpoints, and head gradients are
moved on dedicated `host_to_device` / `device_to_host` CUDA streams that
overlap main-stream compute, via double buffering in `GradientAccumulator.run`.
Ordering in the cell loop uses GPU-side events, so the host thread isn't blocked.

Re-implements the `cpu_gpu2` design on current `main` and fixes the races that
made that branch produce NaN weights.

## Race fixes vs `cpu_gpu2`

- The d2h checkpoint copy now waits on a main-stream event recorded *after* its
  source is produced (on `cpu_gpu2`, layer L+1's input could be copied before
  the `torch.cat` that computes it — the main NaN source).
- One `torch.cuda.synchronize()` at backward start makes forward checkpoint
  writes host-visible before backward reads them.
- Explicit cross-stream events at the start of `run`, replacing reliance on the
  per-row synchronize.

Also skips that per-row `torch.cuda.synchronize()` (host-blocking ~1 s/row) in
the async path, where events already order things. `async_cpu_transfer=False`
leaves all paths unchanged.

## Testing

- New `test/kvcache/test_gradient_async.py`: compares losses/gradients with the
  flag off vs on over two batches, across cache policy × checkpoint
  quantization. 5/5 pass; existing gradient suites green.
- Full-scale Qwen3-4B LoRA (the config that NaN'd on `cpu_gpu2`): 3 clean steps,
  no NaNs, losses matching the async-off control; ~4–8% faster/iter.
Closes #<62>.

----

### DIFF
--- a/keys_values/finetune/args.py
+++ b/keys_values/finetune/args.py
@@ -231,6 +231,10 @@ class GradientArgs:
         cachecp_pin_memory: If `True`, the CPU memory pages for KV cache
             checkpoints are pinned. This can run faster, but also needs more
             real CPU memory.
+        async_cpu_transfer: If `True`, CPU -> GPU and GPU -> CPU transfers
+            during gradient computation are run on separate CUDA streams, in
+            parallel with GPU computation. Requires `layercp_pin_memory=True`
+            and `cachecp_pin_memory=True`. Defaults to `False`.
         debug_print_annotations: If `True`, debug logging during `backward`
             computations are written which allow to track annotations for
             `autograd` saved tensors hooks.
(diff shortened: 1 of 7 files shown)


mseeger:
Thanks! But you said the impact is minor?

I still wonder why that is. Sure, maybe the CPU/GPU transfer is anyway cheap in our case (but I think we could use more of it).

But why is it we see all these events in profiling which seem to block GPU computations?

mseeger:
Sorry, did not look at this for a long while. I'll try to spend more time after the paper deadline.

mseeger (approved):

mseeger:
@vihangp , could you please do two things?
- Rebase and resolve conflicts (I can't somehow do this here)
- Run `./black.sh`

PR #141, opened 2026-07-29
Apply final norm ln_f and logit softcapping before head in LongContextInference

### LINKED ISSUE #139: LongContextInferenceModel omits transformer.ln_f before lm_head (wrong log-probs / first generated token)
**Describe the bug**

`LongContextInferenceModel` does not apply the final norm `transformer.ln_f` before `lm_head`, whereas both `GPT.forward` and the gradient/training path do. As a result, logits (and therefore log-probs and the first generated token) produced through `LongContextInferenceModel` are computed from *un-normalized* final-layer outputs.

Locations (all in `keys_values/long_context.py`, `LongContextInferenceModel._forward_internal_no_check`):

- Scoring / loss path: line ~970 calls `compute_loss_with_limited_logits_tensor(...)`, which at line ~294 does `x = gpt_model.lm_head(model_outputs_for_chunk[...])` — no `ln_f`.
- Generation path: lines ~1000–1002 `logits_final_position = self.gpt_model.lm_head(logits_final_position)` — no `ln_f`.
- The non-logits head branch (line ~979, `compute_loss_for_chunk`) likewise receives raw pre-`ln_f` outputs.

For comparison, the correct behavior is applied elsewhere:
- `keys_values/model.py`, `GPT.forward`: `x = self.transformer.ln_f(x)` then `lm_head(x)`.
- `keys_values/kvcache/gradient/accumulate.py:930`: `model_outputs = gpt_model.transformer.ln_f(x)` then `lm_head(model_outputs)`.

So the training gradient path is correct, but the inference path used by `compute_logprobs` and by generation prefill is not.

**To reproduce**

Minimal, CPU-only, on a **dense (lossless) cache** — where the chunked forward should be exact:

```python
import torch, torch.nn as nn
from keys_values.kvcache.factory import KVCacheFactory
from keys_values.logprobs import compute_logprobs

cfg = Config(n_layer=2, n_head=4, n_embd=64, n_query_groups=2,
             block_size=512, vocab_size=256, padded_vocab_size=256, intermediate_size=128)
torch.manual_seed(0)
with torch.device("cpu"):
    model = GPT(cfg).eval()
torch.manual_seed(1)
seq, k = torch.randint(0, cfg.vocab_size, (1, 24)), 12  # score tokens seq[:, k:]

@torch.no_grad()
def ref_logprobs(m):  # ground truth: plain no-cache causal forward
    m.assign_kv_caches([None] * cfg.n_layer); m.max_seq_length = seq.shape[1]
    lp = torch.log_softmax(m(seq), dim=-1)
    return torch.stack([lp[:, k+j-1].gather(-1, seq[:, k+j].view(-1, 1)).squeeze(-1)
                        for j in range(seq.shape[1]-k)], dim=1)

ref = ref_logprobs(model)
model.assign_kv_caches(KVCacheFactory.create(gpt_model=model, name="dense-default",
    max_batch_size=1, cache_length=24, dtype=torch.float32, cache_kwargs={}))
got, _ = compute_logprobs(model, input_ids=seq[:, :-1], targets=seq[:, k:], chunk_size=1000)
print("max|Δ| compute_logprobs vs no-cache ref: ", (got - ref).abs().max().item())  # ~0.13

# Removing ln_f from the reference makes them match -> ln_f is the cause:
model.assign_kv_caches([None] * cfg.n_layer)
model.transformer.ln_f = nn.Identity()
print("max|Δ| compute_logprobs vs ref WITHOUT ln_f:", (got - ref_logprobs(model)).abs().max().item())  # ~5e-7
```

Output:

```
max|Δ| compute_logprobs vs no-cache ref:  0.1339...
max|Δ| compute_logprobs vs ref WITHOUT ln_f: 4.77e-07
```

Additional isolation performed:
- A direct `model(seq[:, :-1])` call with the same dense cache matches the no-cache reference exactly (0.0) — so the KV cache and attention are fine.
- Single-token decode (`gpt_model(token)`, i.e. `GPT.forward`) matches the reference exactly, except the very first generated token, which is sampled from the `LongContextInferenceModel` prefill logits (missing `ln_f`).
- The discrepancy is independent of `chunk_size` (`chunk=4` and a single big chunk give identical results) and reproduces on a dense cache, so it is not chunking- or eviction-related.

**Expected behavior**

Log-probs / logits from `LongContextInferenceModel` should equal a plain causal `GPT.forward` (they do once `ln_f` is applied). On a dense cache they should be exact.

**Impact**

- `keys_values.logprobs.compute_logprobs` returns incorrect log-probs at every position. This affects RL/GRPO old-log-prob scoring (the `rescore` path) and any other direct use.
- `batched_generate_fn` / `generate_fn` sample the first token of each sequence from un-normalized logits (subsequent tokens use `GPT.forward` and are correct). This feeds the evaluation harness (`SampleBasedMetricsEvaluator`).
- The magnitude is small on a random 2-layer model but scales with how much `ln_f` changes activations, so it can be substantial on trained models.

**Suggested fix**

Apply `self.gpt_model.transformer.ln_f(...)` to the final-layer outputs before the head in `_forward_internal_no_check` — for the `needs_logits` branch, the non-logits branch, and `logits_final_position` — mirroring `GPT.forward` and `kvcache/gradient/accumulate.py:930`.

**Notes**

`git log -S "ln_f" -- keys_values/long_context.py` is empty: `ln_f` has never appeared in that file, so this has been present since the file's initial revision. Environment: CPU, fp32, `dense-default` KV cache.

vihangp:
LongContextInferenceModel._forward_internal_no_check fed raw final-layer outputs into the head, omitting transformer.ln_f (and logit softcapping), whereas GPT.forward and the gradient path in kvcache/gradient/accumulate.py apply both. This made log-probs from compute_logprobs, prefill logits for generation, and loss values reported during training/validation incorrect. Gradients were unaffected, since the backward path applies ln_f correctly.

ln_f is applied in the head/loss computation only, after layer input checkpointing, since checkpoints must store pre-norm outputs (the backward path applies ln_f itself when re-reading them).

Fixes #139

> REPLACE THIS TEXT BLOCK
>
> Describe the reason for this change, what the solution is, and any
> important design decisions you made. 
>
> Remember to follow the [CONTRIBUTING GUIDE] for any code you submit.
>
> [CONTRIBUTING GUIDE]: https://github.com/awslabs/keys_values/blob/main/CONTRIBUTING.md

Closes #<issue number here>.

----

### DIFF
--- a/test/test_long_context.py
+++ b/test/test_long_context.py
@@ -1,0 +1,200 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License").
+# You may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+Tests for :class:`keys_values.long_context.LongContextInferenceModel`.
+
+On a dense (lossless) KV cache, the chunked forward must be exactly
+equivalent to a plain causal `GPT.forward`. In particular, the final
+layer norm `transformer.ln_f` and logit softcapping must be applied
+before the head, see
+https://github.com/awslabs/keys_values/issues/139
+"""
+
+from typing import Optional
+
+import pytest
+import torch
+
+from keys_values.config import Config
+from keys_values.head_model import CrossEntropyOnLogits, HeadModel
+from keys_values.kvcache.factory import KVCacheFactory
+from keys_values.long_context import LongContextInferenceModel
+from keys_values.model import GPT
+
+SEQ_LENGTH = 48
+NUM_TARGETS = 16
+
+
+def _small_config(final_logit_softcapping: Optional[float] = None) -> Config:
+    return Config(
+        n_layer=2,
+        n_head=4,
+        n_embd=64,
+        n_query_groups=2,
+        block_size=512,
+        vocab_size=256,
+        padded_vocab_size=256,
+        intermediate_size=128,
+        final_logit_softcapping=final_logit_softcapping,
+    )
+
+
+def _make_model_and_data(config: Config):
+    torch.random.manual_seed(31415927)
+    with torch.device("cpu"):
+        gpt_model = GPT(config)
+        gpt_model.apply(gpt_model._init_weights)
+    gpt_model.eval()
+    token_ids = torch.randint(0, config.vocab_size, (2, SEQ_LENGTH))
+    return gpt_model, token_ids
+
+
+def _plain_forward(gpt_model: GPT, input_ids: torch.Tensor) -> torch.Tensor:
+    """Reference: plain causal forward without KV caches."""
+    gpt_model.assign_kv_caches([None] * gpt_model.config.n_layer)
+    gpt_model.max_seq_length = input_ids.shape[1]
+    with torch.no_grad():
+        return gpt_model(input_ids)
+
+
+def _assign_dense_caches(gpt_model: GPT, batch_size: int, cache_length: int):
+    gpt_model.assign_kv_caches(
+        KVCacheFactory.create(
+            gpt_model=gpt_model,
+            name="dense-default",
+            max_batch_size=batch_size,
+            cache_length=cache_length,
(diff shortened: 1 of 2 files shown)


test/test_long_context.py
```diff
@@ -0,0 +1,200 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
mseeger, line 1: Thanks for the fix, and especially for the new test.

I'll merge this. As a second step, I'd like to refactor this so that `GradientAccumulator.run_head_model` and `LongContextInferenceModel._forward_internal_no_check` call the same internal code. But I do this as another PR.

mseeger (approved):

PR #142, opened 2026-07-30
Safe eviction defaults (#140), GRPO single-epoch cache reuse, HELMET GRPO driver

### LINKED ISSUE #140: Evicting KV cache corrupts generation (garbage output once prompt > cache_length)
**Describe the bug**

Autoregressive generation through an **evicting** KV cache (e.g. `lastrec`, `h2o`,
and quantized variants) produces garbage — the model emits EOS/newlines
immediately — whenever the prompt is longer than `cache_length` (so the prefill
spans multiple chunks and eviction occurs). With a cache large enough to hold the
whole prompt (no eviction), generation is correct and matches a plain dense /
no-cache forward. This is a **cliff**, not gradual lossy degradation, and it fires
even when only *irrelevant* tokens are evicted, so it is a mechanism bug rather
than expected sparse-attention information loss.

Since eviction is the whole point of these caches for long context (and eval
normally uses `cache_length < sequence_length`), this affects long-context
generation/eval broadly.

**Evidence (Qwen2.5-0.5B-Instruct, A10G, bf16, eager SDPA)**

Controlled key-value retrieval where the answer is the **last** record (so
`lastrec`, which keeps the most-recent tokens, provably retains the needle *and*
the question), ~2062-token prompts, accuracy = substring match of the value:

| cache_length | eviction? | accuracy |
|---:|:--:|---:|
| 2112 | no  | **0.90** |
| 1900 | yes (drops only first ~162 filler tokens) | **0.00** |
| 1536 | yes | 0.00 |
| 1024 | yes | 0.00 |
| 512  | yes | 0.00 |

- The collapse happens at the *first* eviction, dropping only early filler tokens
  that are irrelevant to a last-record query — inconsistent with legitimate
  lossiness.
- Both `lastrec-default` and `lastrec-torch-quantized8` collapse to 0.00, so it is
  **not** a quantization issue.
- `h2o-torch-quantized8` behaves identically.

**Localization**

Only the first prefill chunk (`input_pos == 0`) takes the `_prefill` path
(attends to itself). Every subsequent chunk — which is where eviction happens —
goes through the `_forward` path and uses `token_positions()` to build the
attention mask (`build_mask_slice` in `keys_values/attention/base.py`). The bug
requires a multi-chunk prefill, so it is in that shared masked-attention /
eviction path. `token_positions()` itself appears to return correct original
positions, which points at the mask-slice / eviction-slot bookkeeping. The
first *generated* token (straight from the prefill final-position logits) is
already wrong, so the corruption is in the prefill, before any decode step.

**Severity is scale-dependent (why it may slip past tests):** a tiny 2-layer
random model in fp32 shows only a mild final-logit perturbation under eviction
(max|Δ| ≈ 0.08, argmax preserved), whereas the real 24-layer model in bf16
collapses to garbage. Unit tests on small models may therefore pass.

**To reproduce**

```python
import torch, lightning as L
from litgpt.prompts import PromptStyle, has_prompt_style, load_prompt_style
from litgpt.tokenizer import Tokenizer
from litgpt.utils import auto_download_checkpoint, load_checkpoint
from keys_values.data.constants import LIT_MODEL_FNAME
from keys_values.kvcache.factory import KVCacheFactory, deallocate_kv_cache_buffers_of_model
from keys_values.long_context import LongContextInferenceModel
from keys_values.generate.base import batched_generate_fn
from keys_values.utils import VerbosityLevels

fab = L.Fabric(devices=1, accelerator="cuda", precision="bf16-true")
cd = auto_download_checkpoint(model_name="Qwen/Qwen2.5-0.5B-Instruct")
tok = Tokenizer(cd); cfg = Config.from_file(cd / "model_config.yaml")
ps = load_prompt_style(cd) if has_prompt_style(cd) else PromptStyle.from_config(cfg)
pad = tok.processor.token_to_id("<|endoftext|>") or int(tok.eos_id)
with fab.init_module(empty_init=True): m = GPT(cfg)
load_checkpoint(fab, m, cd / LIT_MODEL_FNAME); m.to(fab.device)

# ~2000-token prompt of "key is value" records; the ANSWER is the LAST record.
import random; rng = random.Random(0)
recs = [f"The value for key '{''.join(rng.choice('abcdefghij') for _ in range(6))}' is {rng.randint(100000,999999)}." for _ in range(180)]
answer = "424242"; recs.append(f"The value for key 'zzzzzz' is {answer}.")
body = ("Records:\n" + " ".join(recs) +
        "\n\nQuestion: what is the value for key 'zzzzzz'? Reply with only the value.\nAnswer:")
ids = tok.encode(ps.apply(body), device=fab.device).unsqueeze(0)
L_ = ids.shape[1]

def gen(cache_len):
    deallocate_kv_cache_buffers_of_model(m)
    m.assign_kv_caches(KVCacheFactory.create(gpt_model=m, name="lastrec-default",
        max_batch_size=1, cache_length=cache_len, dtype=torch.bfloat16))
    m.eval(); m.max_seq_length = L_ + 16
    inf = LongContextInferenceModel(m, head_model=None, chunk_size=min(512, cache_len),
                                    verbose=VerbosityLevels.NONE)
    out = list(batched_generate_fn(model=inf, prompts=ids, max_returned_tokens=8,
        sample_args=dict(temperature=1.0, top_k=1, top_p=1.0), stop_tokens=(), ignore_index=pad))
    return tok.decode(torch.cat(out, dim=1)[0])

print("prompt_len", L_, "answer", answer)
print("no-evict (cl=L+32):", repr(gen(L_ + 32)))   # -> '424242'
print("evict   (cl=1024) :", repr(gen(1024)))       # -> garbage (e.g. '\n\n\n...')
```

Expected: both should return `424242` (the last record is retained under
`lastrec` either way). Actual: the no-eviction case is correct; the eviction case
returns garbage.

**Environment**: Ubuntu 24.04, A10G, torch 2.13.0+cu130, `lastrec` / `h2o` caches,
eager SDPA (FlashInfer disabled to rule the kernel out). Reproduced on the
`grpo-cache-reuse` branch of `mmjerge/keys_values`.

**Related**: possibly the same underlying cause as #99 (`smart-lastrec`) and #5
(`h2o`/`qh2o` scoring); distinct from #139 (`ln_f`).

mmjerge:
## Summary

Fixes #140 (safe eviction defaults), improves the standalone GRPO loop from #128
(single-epoch cache reuse, completion mask, an alignment fix, and a GPU
inference-mode fix), adds `sm_86` to the FlashInfer build, and adds a GRPO
training driver for HELMET tasks plus profiling tooling.

> **Note on the first commit:** `99c6369` is #141 (@vihangp) applied verbatim —
> included so this branch is self-consistent and testable (the GRPO
> measurements depend on it). Once #141 merges, a rebase makes that commit
> vanish; happy to rebase on request.

## Fix #140: safe eviction defaults (attention sinks + grace period)

Evicting KV caches produced garbage generation (immediate EOS) on real models
whenever the prompt exceeded `cache_length`. Instrumentation showed the
eviction/attention mechanics are exact (cached keys/values match a dense
reference gathered at `token_positions()` exactly; the attention output matches
a manual masked-softmax reference to ~1e-7; a 1-layer model shows no divergence
at all). The cause is **attention-sink eviction**: trained LLMs put large
attention mass on the first tokens (BOS / template header) and use them as
no-op sinks (Xiao et al., StreamingLLM); evicting them collapses generation.
Toy random models have no sink structure, which is why unit tests pass.

`KVCacheFactory` now injects safe defaults unless the caller specifies them
(explicit values, including `0`, are respected; `test_utils.create_kv_cache`
pins them to `0` so unit tests keep exercising raw mechanics):

- `lastrec`: `init_grace_tokens = min(16, cache_length // 8)`
- `h2o` / `h2o-vlen` / `qh2o` / `qh2o-vlen`: `grace_period = cache_length // 16`
  (`h2o-orig` excluded — it rejects `grace_period`)

Measured (Qwen2.5-0.5B-Instruct, A10G, bf16, ~2k-token prompts, n=40,
substring exact match; before the fix every evicting config scored 0.000):

| task | dense (full) | h2o @1024 | lastrec @512 | lastrec @1024 |
|---|---:|---:|---:|---:|
| recency (answer = last record) | 0.925 | 0.875 | 0.900 | 0.900 |
| needle_first | 0.700 | 0.200 | 0.000 | 0.000 |
| needle_last | 0.550 | 0.150 | 0.125 | 0.225 |

Recency recovers to within noise of dense — generation through an evicting
cache is functional again — while the needle rows now reflect genuine eviction
lossiness (the expected quality frontier) instead of the collapse.

Grace slots reduce a cache's forward capacity, so chunk sizes are capped at
`cache_length - grace` where applicable (`rl/grpo/loop.py`, examples).

## GRPO improvements (on top of #128)

- **Single-epoch cache reuse**: `GRPOLossHeadModel.set_batch` makes
  `old_logps` optional; when omitted (single-epoch GRPO), the old log-prob is
  `policy_logp.detach()` from the gradient pass's own forward, so the
  importance ratio is exactly 1 and the separate scoring forward pass is
  eliminated (mirrors TRL's `num_iterations=1` behaviour). `grpo_step` gains
  `rescore_old_logps` (multi-epoch / A-B) and `profile` (per-phase timings).
- **Alignment fix**: scoring/gradient passes receive `full_ids[:, :-1]`, so
  completion token `p` is scored against the logits at `p-1` (was off by one).
- **Completion mask**: padding after an early stop no longer contributes to
  the loss or its normalization.
- **GPU inference-mode fix**: cache buffers lazily move to the compute device
  on the first forward; under `torch.inference_mode()` (rollout or eval
  generation) they became inference tensors that the training pass could not
  update in place ("Inplace update to inference tensor outside
  InferenceMode"). `batched_generate_fn` / `generate_completions` gain
  `no_inference_mode` and the GRPO paths use it; eval/generation-only callers
  keep the default (faster) inference mode. GPU-only failure; on CPU buffers
  are already on-device.
- With #141 applied, the rollout-vs-training-forward log-prob skew on a real
  model drops from ~4.8 to ~0.13 (bf16 noise).

## Tooling

- `build_ext.py`: add `sm_86` (A10G / A10 / RTX 30xx) + PTX fallback; kernels
  previously targeted only `sm_80` / `sm_90`.
- `examples/grpo_helmet.py`: GRPO training on HELMET datasets via
  `load_helmet_dev_eval`, reward = `sub_exact_match` (same metric as the eval
  harness), periodic held-out eval, JSONL metrics. Validated end-to-end on
  HELMET `nq` @ 8k (prompts ~7k tokens, H2O cache 4096, real eviction).
- `examples/grpo_profile.py` (completion-length sweep, optimized-vs-rescore),
  `examples/grpo_context_sweep.py` (dense-vs-H2O memory/time across context;
  on a 24 GB A10G dense OOMs at >=24k while H2O stays ~flat, see
  `docs/GRPO_CONTEXT_SCALING.md`), `examples/longqa_task.py` /
  `longqa_quality.py` (controlled eviction-quality probes used for #140).

## Testing

- Full `test/kvcache` suite: 956 passed (includes the factory-default change;
  `test_utils` pins grace to 0 so existing eviction-mechanics tests are
  unchanged).
- `test/rl` (from #128) passes unchanged — the `set_batch` change is
  backward-compatible.
- `test/test_long_context.py` (from #141), `test/generate/test_base.py`,
  `test/finetune`/gradient suites pass.
- GPU (A10G): eviction-fix acceptance table above; end-to-end GRPO training on
  HELMET `nq` with an evicting H2O cache runs cleanly (~13 s/step at ~7k-token
  prompts, 8.2 GB peak).

----

### DIFF
--- a/build_ext.py
+++ b/build_ext.py
@@ -91,8 +91,12 @@ def main():
                     "-U__CUDA_NO_HALF_CONVERSIONS__",
                     "-U__CUDA_NO_BFLOAT16_CONVERSIONS__",
                     "-U__CUDA_NO_HALF2_OPERATORS__",
-                    "-gencode=arch=compute_80,code=sm_80",
-                    "-gencode=arch=compute_90,code=sm_90",
+                    # SASS for the common data-center / workstation GPUs, plus
+                    # PTX for forward-compatible JIT on newer architectures.
+                    "-gencode=arch=compute_80,code=sm_80",  # A100
+                    "-gencode=arch=compute_86,code=sm_86",  # A10G / A10 / RTX 30xx
+                    "-gencode=arch=compute_90,code=sm_90",  # H100
+                    "-gencode=arch=compute_90,code=compute_90",  # PTX (JIT fallback)
                 ],
             },
         )
--- a/build_ext.py
+++ b/build_ext.py
@@ -91,12 +91,8 @@ def main():
                     "-U__CUDA_NO_HALF_CONVERSIONS__",
                     "-U__CUDA_NO_BFLOAT16_CONVERSIONS__",
                     "-U__CUDA_NO_HALF2_OPERATORS__",
-                    # SASS for the common data-center / workstation GPUs, plus
-                    # PTX for forward-compatible JIT on newer architectures.
-                    "-gencode=arch=compute_80,code=sm_80",  # A100
-                    "-gencode=arch=compute_86,code=sm_86",  # A10G / A10 / RTX 30xx
-                    "-gencode=arch=compute_90,code=sm_90",  # H100
-                    "-gencode=arch=compute_90,code=compute_90",  # PTX (JIT fallback)
+                    "-gencode=arch=compute_80,code=sm_80",
+                    "-gencode=arch=compute_90,code=sm_90",
                 ],
             },
         )
--- a/keys_values/generate/base.py
+++ b/keys_values/generate/base.py
@@ -75,37 +75,6 @@ def batched_sample(
     )
 
 
-def _sampled_token_logprobs(
-    logits_stack: torch.Tensor,
-    tokens: torch.Tensor,
-    sample_args_list: list[dict],
-) -> torch.Tensor:
-    """Log-probs of the sampled ``tokens`` under the (temperature-scaled) policy.
-
-    Uses the *full* softmax over the vocabulary (no top-k/top-p truncation), so
-    the value matches what :class:`keys_values.logprobs.LogProbsHeadModel` and
-    :class:`keys_values.rl.grpo.loss.GRPOLossHeadModel` compute during scoring
-    and the policy gradient. This makes the captured log-prob a drop-in
-    replacement for a separate scoring forward pass.
-
-    Args:
-        logits_stack: Model logits, shape ``(batch_size, 1, vocab_size)``.
-        tokens: Sampled tokens, shape ``(batch_size, 1)``.
-        sample_args_list: Per-batch sampling kwargs (for ``temperature``).
-
-    Returns:
-        Log-probs, shape ``(batch_size,)``.
-    """
-    logits = logits_stack[:, -1, :]  # (batch_size, vocab_size)
-    temps = torch.tensor(
-        [max(float(sa.get("temperature", 1.0)), 1e-5) for sa in sample_args_list],
-        device=logits.device,
-        dtype=logits.dtype,
-    ).unsqueeze(-1)  # (batch_size, 1)
-    logp_all = torch.log_softmax(logits / temps, dim=-1)
-    return logp_all.gather(-1, tokens.view(-1, 1)).squeeze(-1)
-
-
 def batched_next_token(
     gpt_model: GPT,
     x: torch.Tensor,
@@ -395,7 +364,11 @@ def _batched_generate_impl(
         int_tokens = [token.item() for token in tokens]
 
         if return_logprobs:
-            step_logps = _sampled_token_logprobs(logits_stack, tokens, sample_args)
+            # RL-specific log-prob capture lives with the rollout code; import
+            # lazily to avoid a generate <-> rl import cycle.
+            from keys_values.rl.grpo.rollout import sampled_token_logprobs
+
+            step_logps = sampled_token_logprobs(logits_stack, tokens, sample_args)
 
         # Check for stop sequences
         stop_dims = []
--- a/keys_values/generate/base.py
+++ b/keys_values/generate/base.py
@@ -75,6 +75,37 @@ def batched_sample(
     )
 
 
+def _sampled_token_logprobs(
+    logits_stack: torch.Tensor,
+    tokens: torch.Tensor,
+    sample_args_list: list[dict],
+) -> torch.Tensor:
+    """Log-probs of the sampled ``tokens`` under the (temperature-scaled) policy.
+
+    Uses the *full* softmax over the vocabulary (no top-k/top-p truncation), so
+    the value matches what :class:`keys_values.logprobs.LogProbsHeadModel` and
+    :class:`keys_values.rl.grpo.loss.GRPOLossHeadModel` compute during scoring
+    and the policy gradient. This makes the captured log-prob a drop-in
+    replacement for a separate scoring forward pass.
+
+    Args:
+        logits_stack: Model logits, shape ``(batch_size, 1, vocab_size)``.
+        tokens: Sampled tokens, shape ``(batch_size, 1)``.
+        sample_args_list: Per-batch sampling kwargs (for ``temperature``).
+
+    Returns:
+        Log-probs, shape ``(batch_size,)``.
+    """
+    logits = logits_stack[:, -1, :]  # (batch_size, vocab_size)
+    temps = torch.tensor(
+        [max(float(sa.get("temperature", 1.0)), 1e-5) for sa in sample_args_list],
+        device=logits.device,
+        dtype=logits.dtype,
+    ).unsqueeze(-1)  # (batch_size, 1)
+    logp_all = torch.log_softmax(logits / temps, dim=-1)
+    return logp_all.gather(-1, tokens.view(-1, 1)).squeeze(-1)
+
+
 def batched_next_token(
     gpt_model: GPT,
     x: torch.Tensor,
@@ -216,7 +247,6 @@ def generate_fn(
         deallocate_kv_cache_buffers_of_model(model.gpt_model)
 
 
-@torch.inference_mode()
 def batched_generate_fn(
     model: LongContextInferenceModel,
     prompts: torch.Tensor,
@@ -226,6 +256,45 @@ def batched_generate_fn(
     sample_args: Union[list[dict], dict],
     stop_tokens: Tuple[List[int], ...] = (),
     deallocate_cache_buffers: bool = True,
+    return_logprobs: bool = False,
+    no_inference_mode: bool = False,
+) -> Iterator[torch.Tensor]:
+    """Generate tokens for a batch of prompts (see :func:`_batched_generate_impl`).
+
+    ``no_inference_mode``: when ``True``, run under ``torch.no_grad`` instead of
+    ``torch.inference_mode``. Required when the KV-cache buffers written during
+    generation are subsequently updated in place by a gradient pass (RL
+    rollouts). Tensors created or moved under ``inference_mode`` -- e.g. the
+    lazy device conversion of cache buffers in ``KVCache._convert_or_check_device``
+    on the first forward -- become "inference tensors" that cannot be modified
+    in place outside inference mode, which breaks the following training
+    forward/backward. Eval and generation-only callers keep the default (faster)
+    inference mode.
+    """
+    ctx = torch.no_grad() if no_inference_mode else torch.inference_mode()
+    with ctx:
+        yield from _batched_generate_impl(
+            model,
+            prompts,
+            max_returned_tokens,
+            ignore_index=ignore_index,
+            sample_args=sample_args,
+            stop_tokens=stop_tokens,
+            deallocate_cache_buffers=deallocate_cache_buffers,
+            return_logprobs=return_logprobs,
+        )
+
+
+def _batched_generate_impl(
+    model: LongContextInferenceModel,
+    prompts: torch.Tensor,
+    max_returned_tokens: int,
+    *,
+    ignore_index: int = -100,
+    sample_args: Union[list[dict], dict],
+    stop_tokens: Tuple[List[int], ...] = (),
+    deallocate_cache_buffers: bool = True,
+    return_logprobs: bool = False,
 ) -> Iterator[torch.Tensor]:
     """
     Generates tokens for a batch of prompts.
@@ -249,11 +318,23 @@ def batched_generate_fn(
             returned. They'll be followed by `ignore_index`.
         deallocate_cache_buffers: Whether to deallocate KV cache buffers at
             the end.
+        return_logprobs: If ``True``, each yielded item is a 3-tuple
+            ``(tokens, logprobs, active_mask)`` instead of just ``tokens``.
+            ``logprobs`` are the per-token log-probs of the sampled tokens
+            under the temperature-scaled policy (full softmax, no top-k/top-p),
+            shape ``(batch_size, 1)``, zeroed for already-stopped rows.
+            ``active_mask`` is a bool tensor, shape ``(batch_size, 1)``, ``True``
+            for rows that were still generating at this step (i.e. real
+            tokens, including the one that triggers a stop sequence). This lets
+            a GRPO rollout capture old-policy log-probs without a separate
+            scoring pass.
 
     Yields:
         Tensors of shape `(batch_size, num)`, where `num >= 1`. Usually,
         `num == 1` (single tokens). The entries for batch dimensions where
-        generation has stopped, are set to `ignore_index`.
+        generation has stopped, are set to `ignore_index`. If
+        ``return_logprobs`` is ``True``, yields ``(tokens, logprobs,
+        active_mask)`` tuples instead.
 
     """
     if prompts.ndim == 1:
@@ -306,15 +387,15 @@ def batched_generate_fn(
     tokens = None
     for current_idx in range(max_returned_tokens):
         if current_idx == 0:
-            tokens = batched_sample(logits_final_position, kwargs=sample_args)
+            logits_stack = logits_final_position
             logits_final_position = None
         else:
-            tokens = batched_next_token(
-                gpt_model=gpt_model,
-                x=tokens,
-                kwargs=sample_args,
-            )
+            logits_stack = gpt_model(tokens)
+        tokens = batched_sample(logits_stack, kwargs=sample_args)
         int_tokens = [token.item() for token in tokens]
+
+        if return_logprobs:
+            step_logps = _sampled_token_logprobs(logits_stack, tokens, sample_args)
 
         # Check for stop sequences
         stop_dims = []
@@ -336,7 +417,20 @@ def batched_generate_fn(
                     seq_pos = int(seq_pos > 0 and int_token == seq[0])
                 stop_progresses[batch_idx][seq_idx] = seq_pos
 
-        yield torch.where(stopped_mask, ignore_ind_vec, tokens.flatten()).view(-1, 1)
+        token_out = torch.where(
+            stopped_mask, ignore_ind_vec, tokens.flatten()
+        ).view(-1, 1)
+        if return_logprobs:
+            # `stopped_mask` reflects rows that stopped in a *previous* step,
+            # so `~stopped_mask` marks the tokens generated at this step (the
+            # stop-triggering token included; its stop flag is set below).
+            active = ~stopped_mask
+            logp_out = torch.where(
+                stopped_mask, torch.zeros_like(step_logps), step_logps
+            ).view(-1, 1)
+            yield token_out, logp_out, active.view(-1, 1)
+        else:
+            yield token_out
         for batch_idx in stop_dims:
             has_stopped[batch_idx] = True
             stopped_mask[batch_idx] = True
--- a/docs/GRPO_CONTEXT_SCALING.md
+++ b/docs/GRPO_CONTEXT_SCALING.md
@@ -1,0 +1,84 @@
+# GRPO context scaling: dense vs. H2O (sparse) KV cache
+
+Measures what the KeysAndValues sparse KV cache buys for GRPO as prompt/context
+length grows: full attention (`dense-default`, cache holds every token) vs. H2O
+(`h2o-torch-quantized8`, cache capped at a fixed budget so it evicts once the
+prompt exceeds it).
+
+Reproduce with `examples/grpo_context_sweep.py`.
+
+## Setup
+
+- **Hardware**: AWS g5.8xlarge, 1× NVIDIA A10G (24 GB), 124 GB RAM, driver 595, CUDA 13.2.
+- **Model**: Qwen2.5-3B-Instruct (bf16), GQA group_size 8 (FlashInfer-compatible).
+- **Attention kernels**: vendored FlashInfer SDPA (built for sm_86), returning
+  attention weights for H2O scoring.
+- **GRPO step**: batch 2 (1 prompt × group 2), `max_new_tokens=32`,
+  `chunk_size=1024`, `layers_per_cell=1`, single-epoch (detach) path.
+- **H2O cache budget**: 4096 slots (fixed). Dense cache = full sequence length.
+- **Timing**: 1 warmup + 3 measured steps per config; per-phase wall-clock (ms),
+  CUDA-synchronized. Peak = `torch.cuda.max_memory_allocated`.
+
+## Results
+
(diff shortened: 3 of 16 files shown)


mmjerge:
## End-to-end results: GRPO on HELMET through an evicting sparse cache

Ran the full pipeline in this PR on HELMET `nq` (8k bucket; real RAG-QA prompts, ~7k tokens avg), Qwen2.5-0.5B-Instruct, single A10G (24 GB), bf16, eager SDPA. Two GRPO arms, identical except the KV cache: **H2O @ 4096 slots (actively evicting during rollouts and the gradient pass)** vs **dense (full attention)**. Reward = `sub_exact_match` (round 1) / `max(EM, token-F1)` partial credit (round 2), same metric family as the eval harness.

### Cross-evaluation: training cache × inference cache (n=100 held-out records)

| checkpoint ↓ · eval cache → | dense (EM / F1) | H2O@4096 (EM / F1) |
|---|---:|---:|
| base (untrained) | 0.330 / 0.178 | 0.320 / 0.142 |
| GRPO-trained under H2O | 0.350 / 0.181 | 0.340 / 0.144 |
| GRPO-trained under dense | 0.340 / 0.180 | 0.300 / 0.132 |

**Findings**

1. **Sparse inference is essentially free on this task.** Base model under H2O@4096 vs dense: 0.320 vs 0.330 EM (~1pp, within noise at n=100). With the eviction defaults from this PR in place, a cache at ~55% of the prompt length costs no measurable quality on real long-context QA. (Before the #140 fix, every evicting configuration scored 0.000.)
2. **Training under the sparse cache does not damage the policy.** All checkpoints are within noise of each other under matched eval conditions; the H2O-trained policy is indistinguishable from the dense-trained one.
3. **Cross-eval matters.** Per-arm evals (each arm under its own cache, small n) had suggested a large H2O-vs-dense gap (0.292 vs 0.417); under matched inference and n=100 that gap disappears entirely — it was inference-cache conflation plus small-sample noise.

### Training dynamics / footprint

- GRPO steps through the evicting cache run stably end-to-end (rollout, scoring-free single-epoch path, chunked backward): ~13–24 s/step at ~7k-token prompts on the A10G, ~8 GB peak — roughly half the dense footprint, consistent with `docs/GRPO_CONTEXT_SCALING.md` (dense OOMs at ≥24k ctx on 24 GB while H2O stays flat).
- The F1 partial-credit reward raised the fraction of steps carrying nonzero group-relative advantage ("signal rate") to ~0.7–0.8, from mostly-degenerate groups under binary EM.
- Caveat: at lr 1e-6, ≤300 steps, one prompt × 8 rollouts per update, **neither arm materially improves over base yet** — the runs above validate stability and parity, not learning gains. A hotter run (lr ~5e-6, 500+ steps, accumulated prompts, multiple seeds) is the follow-up for the "H2O tracks dense *improvement*" curve.

Repro: `examples/grpo_helmet.py` (arms) and `examples/grpo_helmet_crosseval.py` (matrix), both in this PR.

mmjerge:
Follow-up: the round-2 arms (F1 partial-credit reward, group size 8, 300 steps) cross-evaluate the same way (n=100):

| checkpoint ↓ · eval cache → | dense (EM / F1) | H2O@4096 (EM / F1) |
|---|---:|---:|
| GRPO-trained under H2O (v2) | 0.340 / 0.180 | 0.330 / 0.143 |
| GRPO-trained under dense (v2) | 0.330 / 0.179 | 0.300 / 0.139 |

Consistent with round 1: all cells within noise of the untrained base (0.330 / 0.320), sparse-inference cost ~1–3pp EM. The parity conclusions hold; the "learning gains" run (higher lr, more steps/prompts, multiple seeds) remains future work.

mmjerge:
## Hot-run results: learning through the evicting cache is real (with caveats)

Second round with lr 5e-6, 8 rollouts x 2 accumulated prompts x 400 updates (gradient accumulation added in this PR), F1 partial-credit reward, single seed, n=100 cross-eval.

**nq @ 8k** (EM / F1):

| checkpoint ↓ · eval → | dense | H2O@4096 |
|---|---:|---:|
| base | 0.330 / 0.178 | 0.320 / 0.142 |
| GRPO-trained under H2O | **0.450** / 0.166 | **0.420** / 0.167 |
| GRPO-trained under dense | 0.270 / **0.397** | 0.220 / 0.396 |

**hotpot_qa @ 8k**:

| checkpoint ↓ · eval → | dense | H2O@4096 |
|---|---:|---:|
| base | 0.290 / 0.145 | 0.180 / 0.109 |
| GRPO-trained under H2O | 0.300 / 0.316 | 0.230 / 0.271 |
| GRPO-trained under dense | **0.410** / **0.436** | 0.320 / 0.390 |

Findings:

1. **Training through the evicting H2O cache produces real gains**: +12pp EM on nq (0.33 -> 0.45), the strongest EM result in the matrix.
2. **Task dependence**: on hotpot the dense arm gains more (+12pp vs +1pp EM), and H2O *inference* costs ~11pp EM on hotpot vs ~1pp on nq -- multi-hop tasks with dispersed evidence are where eviction hurts, both at inference and during training rollouts.
3. **Reward-shaping artifact in one arm**: the dense/nq arm shows train reward rising (0.70) while eval EM collapses (0.44 -> 0.12) and F1 doubles -- the max(EM, token-F1) partial credit appears to have been gamed. Sample inspection is underway; the next round separates the reward (EM + small F1 tiebreaker) from the eval metric.
4. **Statistics**: single seed, n=100 (SE ~5pp); the +/-12pp deltas are ~2-sigma. Seed replications are queued.

mseeger:
@mmjerge I am aware of attention sink. And indeed, when I evaluated a pre-trained checkpoint with the `lastrec` policy, I did not get good results.

This is also why I implemented `smart-lastrec`.

I am OK with have a default grace period with `lastrec` of 1 (or a small number).

But I do not see why this would affect H2O (say). If the model uses the first token for attention, H2O would keep it in the cache (as it is heavily used).

In the pre-trained checkpoint evals, I did not see bad results just because of there being no grace period, at least not for H2O.

mmjerge:
**Correction on the "reward hacking" hypothesis, after inspecting generations + seed-1 replication.**

Sample generations (distinct NQ questions, dense eval cache):

- **base**: verbose scaffolded answers — "The last time the Philadelphia Eagles played the New England Patriots was **in Super Bowl LII**, which took place..."
- **h2o-trained**: same verbose style, largely intact.
- **dense-trained**: *terse* answers — "Super Bowl LII in 2017.", "Mr Carson.", "March 10, 2017."

So the dense arm did not become verbose — the F1 partial credit taught it to strip scaffolding (terse answers score higher token-F1). The EM collapse is substantially a **metric-style artifact**: HELMET NQ targets often include leading prepositions ("**in** Super Bowl LII", "**on** October 21, 2016"), and `sub_exact_match` requires the target as a substring. The base/h2o style ("...was in Super Bowl LII...") contains it; the terse style ("Super Bowl LII in 2017") does not, even when semantically correct. Some genuine wrong answers are also present in all arms.

**Seed-1 replication (nq, n=100)** reproduces the seed-0 pattern: h2o-trained 0.44/0.45 EM (dense/h2o eval) vs base 0.33; dense-trained 0.31/0.29. So the pattern is robust across 2 seeds, but the safe interpretation is:

- **Robust claim**: GRPO through an actively evicting H2O cache trains stably and moves the policy as much as dense training does — at ~half the memory. No evidence sparse-cache training damages learning.
- **Not yet safe**: "H2O-trained beats dense-trained" on EM — the delta is confounded by answer-style/metric interaction (both arms shifted style under the F1 reward; the dense arm shifted further per update).
- **Fix for the next round**: reward = EM + small F1 tiebreaker (e.g. EM + 0.2*F1) so style stays anchored to the eval metric, and report both metrics.

hotpot seed-1 arms are still running; will append when complete.

mmjerge:
@mseeger Agreed — I've replied in detail on #140 (with an isolated example), and updated this PR accordingly:

- The factory comment now states the two mechanisms separately: `init_grace_tokens` for `lastrec` is about **attention sinks** (your point — H2O is not affected by sink eviction, since `keep_initial_fraction` + accumulated scores retain the initial tokens); `grace_period` for the H2O family is about **fresh tokens having the lowest accumulated scores**, i.e. the recently-read tail being evicted under pressure — the case the `H2OKVCache` docstring already anticipates.
- If you'd prefer to keep H2O's `grace_period` default at 0 (no behavior change) and only default `init_grace_tokens` for `lastrec`, I'm happy to trim the PR to that — the measured data for the H2O case is in #140 so the tradeoff is documented either way.

Also pushed since your comment: an `em_f1` reward option for the HELMET driver (EM + 0.2*F1 — keeps the reward anchored to the eval metric after we caught the F1-only reward drifting answer *style*; details in the thread above).

mmjerge:
Rebased onto `main` now that #141 is merged — the carried copy of that commit dropped out as promised; this PR is now only the eviction defaults, GRPO changes, and tooling. Tests re-verified after rebase (test/rl, test_long_context, kvcache).

mmjerge:
**Seed-1 replication complete** (same recipe: lr 5e-6, group 8 x 2 accumulated prompts x 400 updates, F1 reward, n=100 cross-eval). Both seeds agree on the direction of every cell.

**hotpot_qa @ 8k, seed 1** (EM / F1; seed-0 in parentheses):

| checkpoint ↓ · eval → | dense | H2O@4096 |
|---|---:|---:|
| base | 0.290 / 0.145 | 0.180 / 0.109 |
| H2O-trained s1 | 0.280 / 0.122 (0.300 / 0.316) | 0.290 / 0.120 (0.230 / 0.271) |
| dense-trained s1 | **0.380** / 0.381 (0.410 / 0.436) | 0.330 / 0.335 (0.320 / 0.390) |

**Two-seed summary across both tasks** (EM under dense eval, vs base):

| arm | nq | hotpot_qa |
|---|---:|---:|
| H2O-trained | **+12pp / +11pp** (0.45, 0.44) | ~flat (0.30, 0.28) |
| dense-trained | −6pp / −2pp (0.27, 0.31) | **+12pp / +9pp** (0.41, 0.38) |

Consistent, replicated pattern: **each arm learns where its rollouts see the evidence.** On single-hop nq, the H2O arm improves robustly (+11-12pp both seeds) while the dense arm's EM drop is the answer-style artifact documented above. On multi-hop hotpot — where H2O inference itself costs ~11pp (dispersed evidence gets evicted) — the dense arm improves (+9-12pp) while the H2O arm stays ~flat under dense eval (though it does gain under *matched* H2O inference: 0.18 -> 0.23/0.29 across seeds).

Takeaways: (1) GRPO through an actively evicting cache trains stably and can deliver real gains (nq), at ~half the memory; (2) on dispersed-evidence tasks the eviction-induced information loss in rollouts limits the learning signal — the same regime where H2O inference costs accuracy; (3) the F1-only reward's style drift is confirmed across seeds — subsequent runs use the EM-anchored `em_f1` reward added in this PR.

mmjerge:
## Consolidated results: algorithms x caches x tasks (final for this round)

All: Qwen2.5-0.5B-Instruct, HELMET @ 8k (~7k-token prompts), single A10G, 400 updates (x2 accumulated prompts), lr 5e-6, n=100 cross-eval. "H2O" = h2o-torch-quantized8 @ 4096 slots (actively evicting; grace_period=cl/16 set at application level per #140 discussion). EM / F1 shown as dense-eval | H2O-eval.

**nq @ 8k** (base: 0.33/0.18 | 0.32/0.14):

| method (training cache) | dense-eval | H2O-eval |
|---|---:|---:|
| SFT (H2O) | 0.16 / 0.32 | 0.19 / 0.30 |
| SFT (dense) | 0.16 / 0.35 | 0.23 / 0.36 |
| GRPO em_f1 (H2O) | 0.28 / 0.33 | 0.26 / 0.32 |
| GRPO em_f1 (dense) | 0.29 / 0.39 | 0.29 / 0.42 |
| **RLOO em_f1 (H2O)** | **0.43 / 0.15** | **0.42 / 0.12** |

**hotpot_qa @ 8k** (base: 0.29/0.15 | 0.18/0.11):

| method (training cache) | dense-eval | H2O-eval |
|---|---:|---:|
| SFT (H2O) | 0.31 / 0.39 | 0.28 / 0.39 |
| SFT (dense) | 0.37 / 0.43 | 0.31 / 0.40 |
| GRPO em_f1 (H2O) | 0.38 / 0.44 | 0.33 / 0.40 |
| GRPO em_f1 (dense) | **0.42 / 0.46** | 0.39 / 0.43 |

**Findings**

1. **Training through the actively evicting cache works across all three algorithms** (GRPO, RLOO, SFT) — stable end-to-end, ~half the dense memory footprint.
2. **RL > SFT** on both tasks (hotpot EM 0.42 vs 0.37; nq 0.29-0.43 vs 0.16).
3. **Sparse-trained ~ dense-trained**: GRPO H2O arm within 1-4pp EM of dense (nq 0.28 vs 0.29; hotpot 0.38 vs 0.42), and the **RLOO H2O arm is the best nq result overall (0.43 EM)** — a sparse-trained model beating every dense-trained one.
4. **RLOO's nq gain is the most credible EM improvement of the campaign**: unlike earlier arms, its F1 stays at base level (0.15 vs 0.18), i.e. the answer *style* did not drift — it gets more answers right within the base model's own style. (RLOO also showed the highest signal rate, 0.93-0.94: leave-one-out advantages don't divide by group std, so small reward spreads still carry gradient.)
5. **Training shrinks the sparse-inference penalty**: base loses 11pp EM under H2O eval on hotpot; trained models lose 3-5pp.
6. Caveats: nq's substring-EM remains style-sensitive (SFT/GRPO F1-heavy arms drop nq EM while F1 rises), single seed for round-3/SFT/RLOO (the round-2 pattern replicated across 2 seeds), 0.5B model, 2 tasks.

RLOO mode and the SFT driver live on stacked branches (`rl-rloo`, `sft-helmet`) to keep this PR's scope as reviewed; happy to open follow-up PRs once this one lands.

mseeger:
Hello, I am a bit confused about this PR. Maybe it mixes two things?

- Change of certain defaults in order to avoid poor behavior of `lastrec`
- New code for the RL effort

Could you split the two? As I said, I'd like to be conservative on the defaults, but I am more than happy with new code on the RL front, if this helps.

mseeger:
Please also try to document findings in markdown files checked in. The PR messages will be lost in the end, or very hard to find.

mmjerge:
@mseeger Both points addressed:

1. **Split done.** The eviction defaults now live in #144 (lastrec-only, conservative scope as you preferred); this PR no longer touches `factory.py` / `test_utils.py` and is RL-only (GRPO loop/loss/rollout, generation support, HELMET drivers, docs). The two PRs are independent — the RL drivers set grace at application level, so neither depends on the other's merge order.

2. **Findings checked in.** `docs/GRPO_HELMET_RESULTS.md` (in this PR) consolidates the whole campaign from the thread above: the round-1 parity cross-eval, the round-2 learning gains with the answer-style/metric artifact analysis and seed replication, the round-3 algorithms × caches × tasks matrix (SFT/GRPO/RLOO), caveats, and repro commands.

Re-verified after the split: `test/rl` (19 passed), `test/test_long_context.py` + `test/generate` (17 passed), `test/kvcache` (956 passed on #144).

keys_values/generate/base.py
```diff
@@ -75,6 +75,37 @@ def batched_sample(
     )
 
 
+def _sampled_token_logprobs(
```
mseeger: If this is specific to the RL work, it should move to `keys_values/rl`. Otherwise, you'd have to argue why it is more generally needed (I do not get it from the docstring).

build_ext.py
```diff
@@ -91,8 +91,12 @@ def main():
                     "-U__CUDA_NO_HALF_CONVERSIONS__",
                     "-U__CUDA_NO_BFLOAT16_CONVERSIONS__",
                     "-U__CUDA_NO_HALF2_OPERATORS__",
-                    "-gencode=arch=compute_80,code=sm_80",
-                    "-gencode=arch=compute_90,code=sm_90",
+                    # SASS for the common data-center / workstation GPUs, plus
```
mseeger: Can you explain why this change is needed? This is a script needed to build the FlashInfer stuff by Vihang.
- Do you need this for the different architectures?
- If so, did you test this on these?
- This will not break the current setup?

keys_values/generate/base.py
mmjerge: Agreed — it's RL-specific. Moved it to `keys_values/rl/grpo/rollout.py` as `sampled_token_logprobs` (next to the rollout code that captures old-policy log-probs). The generic generation loop lazy-imports it inside the `return_logprobs` branch to avoid a generate ↔ rl import cycle. Done in the latest push.

build_ext.py
mmjerge: You're right that this doesn't belong in the RL PR — I've reverted `build_ext.py` to match `main` in the latest push, so this PR no longer touches Vihang's build script.

For context on what it was: I'd added `sm_86` (A10G/A10/RTX-30xx) plus a PTX fallback line, because the A10G is the box I run the GRPO experiments on and the stock `sm_80`+`sm_90` build won't load FlashInfer there. The retained `sm_80`/`sm_90` SASS lines were untouched, so it wouldn't have broken the A100/H100 setup — but it was orthogonal to RL and I hadn't tested it beyond the A10G, so it's better out of scope here. If it's useful I'll open a separate one-line PR against `main` for the A10G arch, with you/Vihang as reviewers.

mseeger:
@mmjerge This is mostly RL code, plus a sensible extension of the inference routine. Could you rebase this and make sure tests pass? I'll merge it then.

PR #143, opened 2026-08-04
fix: align the evaluation metrics with baseline method

Jantory:
## Metric Alignment Fix

In this fix, we did two changes to make it aligned with the baseline evaluation:
- For RAG task evaluation, do the lower case for both the target and response
- For ICL task evaluation, remove the redundant block to assign extra bonus to a specific response format

----


mseeger (approved):
I'll merge this now. Still, would be good to have a statement from Konstantinos whether with these changes, the two evaluation codes are equivalent now

PR #145, opened 2026-08-06
Fix HELMET loader reproducibility: seed dev/eval split and demo sampling

mmjerge:
Two unseeded random sites in `load_helmet_dev_eval.py` made the constructed benchmark differ between machines and cache rebuilds, so absolute results were not comparable across hosts. Found while replicating the #142 evals on a fresh instance: base-model dense-eval EM came out 0.11 vs the 0.33 measured on the original box — same code, same model, different (unseeded) eval split and few-shot demos.

- `load_rag`: the 100 eval questions were drawn with bare `random.sample`; now `random.Random(42)`, consistent with `seed = 42` used elsewhere in the module. The split is drawn once and cached to disk, so any two hosts now build identical dev/eval splits.
- `load_citations`: per-instance demo choice used unseeded `random.sample` inside a `.map`; now hash-seeded per question, mirroring the existing RAG-loader pattern (`demos.shuffle(seed=h)`).

Note: `longtrain/` split caches built before this fix retain their old draw; delete the cache directory to rebuild deterministically. Results measured against pre-fix caches (including the numbers posted on #142) are internally consistent but not comparable across machines — re-baselining under the seeded split is in progress.

Testing: `test/data` passes (9 passed). Related: #142 gains a driver-side fix (question-diverse eval record selection) in the same spirit.

### DIFF
--- a/keys_values/data/load_helmet_dev_eval.py
+++ b/keys_values/data/load_helmet_dev_eval.py
@@ -483,8 +483,12 @@ def load_rag(
     instance_data = load_dataset("json", data_files=instance_path)["train"]
     demon_data = load_dataset("json", data_files=demo_path)["train"]
 
+    # Seeded RNG: the dev/eval question split must be identical across
+    # machines, otherwise results computed against different split caches
+    # are not comparable (the split is drawn once and cached to disk).
+    split_rng = random.Random(42)
     if dataset_key in ["nq", "trivia_qa", "hotpot_qa"]:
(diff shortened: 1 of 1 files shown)


mseeger (approved):
Thanks!

Can you explain why the differences in final performance are so large?

mseeger:
Do you have an idea why the differences are so large?

I did not realize the Helmet datasets have such a large variance. This may affect my results as well.

Issue #31, opened 2026-02-16
Refactor node annotations in autograd_hooks.py


mseeger:
**Is your feature request related to a problem? Please describe.**
Currently, node annotations are defined as a single dataclass, and creation and application are hacked into autograd_hooks.py and the training replay caches. This is not clean and hard to extend.

**Describe the solution you'd like**
Abstraction of node annotation as class, with `create`, `match`,  and `apply` methods. Anything specific to a node annotation is provided there. All existing types are subclasses.

mseeger:
The workflow became even more complex due to #149, which robustified the method in the light of non-deterministic reorderings happening in `autograd`.

mseeger:
We want to come up with a clean design, whereby anything specific to annotation types (e.g., "cat", "scatter", "ext") is moved out of `autograd_hooks.py`. Also, we want to clean up the backward traversal of the KV buffer chains and "parking" of buffers which have not yet been requested.

PR #149, opened 2026-08-21
Fix #148: walk annotation chain when final buffer is 2+ chunks ahead

### LINKED ISSUE #148: Chunked backward fails when a generated region spans 3+ chunks (autograd_hooks annotation adjacency)
## Summary

During GRPO training with **long sampled completions** (up to 2600 generated tokens, `chunk_size=1024`), the chunked backward fails stochastically in the replay unpack:

```
File ".../keys_values/kvcache/gradient/autograd_hooks.py", line 785, in unpack_hook
    x = self._unpack_from_annotation(annotation)
File ".../keys_values/kvcache/gradient/autograd_hooks.py", line 852, in _unpack_from_annotation
    raise ValueError(
ValueError: Annotation scatter-value (20,1): (4, 4, 8192, 128): final chunk_idx = 3, must be in [1, 2]
```

`_unpack_from_annotation` assumes the final buffer for a `(layer, kind)` node is at most **one chunk ahead** of the annotation being unpacked (`final_idx in [chunk_idx, chunk_idx + 1]`). In this failure the final buffer is two chunks ahead (3 vs annotation chunk 1), so the invariant breaks and backward dies.

## Where it shows up (and where it never does)

- **Never** in our QA-style GRPO runs: 27-32k-token prompts but only **32-token completions** (hundreds of runs, zero occurrences).
- **Stochastically** in long-generation runs: html-to-TSV extraction (LongProc) with `max_new_tokens=2600`, eos-terminated (ragged per-row lengths). One seed crashed at ~step 30; a parallel seed with different sampling is past step 100 without issue.

So the trigger involves a *generated* (rather than prefilled) region spanning **3+ backward chunks**, with per-row eos raggedness making it sampling-dependent.

## Config of the failing run

Qwen2.5-7B-Instruct (28 layers, GQA 4 KV heads, head 128), bf16, single L40S.
`h2o-torch-quantized8`, `cache_length=8192`, `grace_period=512`, batch (group) 4,
prompt ~14k tokens, `max_new_tokens=2600`, `chunk_size=1024`, `layers_per_cell=1`,
gradient accumulation 2 micro-steps, RLOO advantages. Failing annotation:
layer 20, `scatter-value`, chunk 1, buffer `(4, 4, 8192, 128)`.

## Repro status (honest)

Not yet minimized. CPU attempts with tiny models (2 layers, `cache_length=64`, completions spanning 3 chunks of 16, eos-ragged rows, quantized-8 buffers, grad accumulation, RLOO) do **not** trigger it, so the adjacency violation needs something extra that I have not isolated -- possibly an interaction between eviction timing (`grace_period`), ragged eos stopping near chunk boundaries, and which chunk a `scatter-value` annotation lands in. I can share the full job log and the exact record/seed on request, and I am happy to run instrumented builds (`debug_print_annotations=True` reaches ~step 30 in about 8h on our hardware).

## Workaround

Setting `chunk_size >= max_new_tokens` (so the generated region spans at most 2 chunks) avoids it; that is what our runs use now. Costs memory (larger backward cells), so it is not a fix for very long generations.

## Why it matters

Any RL/SFT use with long *generated* sequences through `LongContextGradientModel` will hit this eventually (it is sampling-dependent, so it looks like flaky infrastructure until you find the invariant). Short-completion training is unaffected.

mmjerge:
## Summary

Fixes #148 (chunked backward fails when a generated region spans 3+ chunks).

`_unpack_from_annotation` assumed the final buffer for a `(layer, kind)` node is at most one chunk ahead of the annotation being unpacked (`final_idx in [chunk_idx, chunk_idx + 1]`). With long generated regions spanning 3+ backward chunks, autograd's backward for the intermediate chunks can be served without unpacking their `scatter-*`/`cat-*` annotations, so the gap grows beyond one chunk and backward dies with e.g. `ValueError: Annotation scatter-value (20,1): final chunk_idx = 3, must be in [1, 2]`.

## Changes

1. **`autograd_hooks.py`: walk the annotation chain across multi-chunk gaps.** All `scatter`/`cat` annotations are guaranteed to be present in `_packed_arg_for_id` after `_flush_remaining_pack_arguments` (they are inserted under fresh IDs even when unmatched, precisely to keep the reconstruction chain complete). `_unpack_from_annotation` now applies the intermediate annotations early, in descending chunk order, and parks each reconstructed intermediate state (as a copy, since the buffer is mutated in place) in `_id_to_unpacked`, so autograd can still fetch those IDs later. `_find_prior_annotation` is generalized to `_find_chain_annotation(annotation, chunk_idx)`; the legacy one-step ext-before-scatter behavior is preserved unchanged.

2. **`rl/grpo/loop.py`: pass `may_match_twice=may_match_twice_flex_attention_sdpa`.** With the new training replay cache, `ext-*` annotations legitimately match two pack arguments. The finetune path (`may_match_twice_factory`) and `test_gradient.py` both pass this predicate; the GRPO loop did not, so the second save of an ext buffer was left as an unmatched raw pack argument. Autograd's backward for that chunk can then be served from the raw copy without advancing the annotation chain — the likely trigger for the adjacency violation. This is the root-cause fix; change (1) makes the unpack robust even if the chain stalls for another reason.

## Why it only showed with long completions

With at most 2 buffer states per cell the gap can never exceed one chunk. QA-style runs (32-token completions) never hit it; long-generation runs (e.g. 2600-token completions, `chunk_size=1024`) hit it stochastically, dependent on per-row eos raggedness — consistent with the reports in #148.

## Testing

- New regression test `test_unpack_walks_multi_chunk_annotation_chain` (parametrized over chunk gaps 1, 2, 4): builds a ground-truth chain of buffer states linked by `scatter-value` annotations, sets the final buffer several chunks ahead, unpacks the earliest annotation, and verifies (a) correct reconstruction, (b) parked intermediate states are served when their IDs are unpacked later. The gap ≥ 2 cases fail with the exact #148 error before the fix and pass after.
- `test/kvcache/test_autograd_hooks.py`: 5 passed
- `test/kvcache/test_gradient.py`: 10 passed
- `test/kvcache/test_gradient_main.py` + `test/rl`: pass (CUDA-only cases skipped; verified on CPU, macOS)

Not verified: an end-to-end GPU reproduction of the original failing GRPO config (long-generation LongProc/longmath runs). We will requeue the previously crashed 32k jobs with this fix and report back on the issue.

### DIFF
--- a/keys_values/kvcache/gradient/autograd_hooks.py
+++ b/keys_values/kvcache/gradient/autograd_hooks.py
@@ -150,6 +150,20 @@ class PackArgumentAsAnnotation:
 
 
 @dataclass(frozen=True)
+class ParkedBufferState:
+    """
+    Buffer state reconstructed ahead of its unpack request (chain walk, see
+    :meth:`CellComputationAutogradHooks._unpack_from_annotation`). Parked on
+    CPU: parking happens during backward, when device memory is tightest,
+    and a full-size device clone per walked chunk can push a large
+    configuration into OOM.
+    """
+
+    x_cpu: torch.Tensor
+    device: torch.device
+
+
+@dataclass(frozen=True)
 class PackArgumentAsIndex:
     index_3d: torch.Tensor
     final_dim: int
@@ -790,6 +804,8 @@ class CellComputationAutogradHooks(AutogradHooks):
             if idd in self._id_to_unpacked:
                 # Unpacked this one before: Just return it
                 x = self._id_to_unpacked[idd]
+                if isinstance(x, ParkedBufferState):
+                    x = x.x_cpu.to(device=x.device)
                 self._id_counts[idd] -= 1
                 if self._id_counts[idd] == 0:
                     # Not needed anymore:
@@ -931,15 +947,26 @@ class CellComputationAutogradHooks(AutogradHooks):
                         # `annot` was applied earlier than autograd asks for
                         # it. Park the state just reconstructed (as a copy,
                         # since `buffer` is modified in place by subsequent
-                        # steps), so :meth:`unpack_hook` can serve the ID later
-                        parked = buffer.clone()
-                        if park_dtype is not None and parked.dtype != park_dtype:
-                            parked = parked.to(dtype=park_dtype)
+                        # steps), so :meth:`unpack_hook` can serve the ID
+                        # later. Device buffers are parked on CPU to avoid
+                        # OOM during backward.
+                        target_dtype = (
+                            park_dtype if park_dtype is not None else buffer.dtype
+                        )
+                        if buffer.device.type == "cpu":
+                            parked = buffer.detach().clone().to(dtype=target_dtype)
+                        else:
+                            parked = ParkedBufferState(
+                                x_cpu=buffer.detach().to(
+                                    device="cpu", dtype=target_dtype
+                                ),
+                                device=buffer.device,
+                            )
                         self._id_to_unpacked[park_id] = parked
                         if park_id not in self._id_counts:
                             self._id_counts[park_id] = 1
                         if self._debug_test_args:
-                            self._debug_log_args.append((parked.clone(), annot))
+                            self._debug_log_args.append((buffer.clone(), annot))
                 # Sanity check
                 assert (
                     annot.shape == buffer.shape
--- a/keys_values/kvcache/gradient/autograd_hooks.py
+++ b/keys_values/kvcache/gradient/autograd_hooks.py
@@ -520,6 +520,7 @@ class CellComputationAutogradHooks(AutogradHooks):
         self._packed_arg_for_id = None
         self._id_counts = None
         self._id_to_unpacked = None
+        self._orphan_annotation_ids = None
         # ID assigned to next pack hook argument
         self._next_id = None
         self._num_matched_annotations = None
@@ -590,6 +591,10 @@ class CellComputationAutogradHooks(AutogradHooks):
         self._packed_arg_for_id: Dict[int, PackedArgumentType] = dict()
         self._id_counts: Dict[int, int] = dict()
         self._id_to_unpacked: Dict[int, torch.Tensor] = dict()
+        # IDs inserted by :meth:`_flush_remaining_pack_arguments` to keep the
+        # annotation chain complete. Nothing in the autograd graph refers to
+        # them, so states reconstructed for them need not be parked.
+        self._orphan_annotation_ids: Set[int] = set()
         self._next_id = 0
         self._num_matched_annotations = 0
         self._num_comparisons = 0
@@ -625,6 +630,8 @@ class CellComputationAutogradHooks(AutogradHooks):
             self._packed_arg_for_id = None
         self._id_counts = None
         self._id_to_unpacked = None
+        if self._orphan_annotation_ids is not None:
+            self._orphan_annotation_ids.clear()
         self._next_id = None
         self._num_matched_annotations = None
         self._num_comparisons = None
@@ -882,17 +889,18 @@ class CellComputationAutogradHooks(AutogradHooks):
                     idd, prior_annotation = found
                     if self.debug_print_annotations:
                         print(f"--> Doing {str(prior_annotation)} first")
-                    if prior_chunk_idx > chunk_idx:
-                        # Applied early: Remove the entry here, park the
-                        # reconstructed state below
-                        value = self._packed_arg_for_id.pop(idd)
-                        chain.append((prior_annotation, idd, value.target_dtype))
-                    else:
-                        # `prior_chunk_idx == chunk_idx` ("ext-*" case): The
-                        # entry stays in `_packed_arg_for_id`. If its ID is
-                        # unpacked later, the "already done" branch above
-                        # serves it
+                    # The entry is applied early, so it is removed here. We
+                    # park the state it reconstructs (below), unless nothing
+                    # in the autograd graph can ask for this ID: orphan IDs
+                    # are inserted by
+                    # :meth:`_flush_remaining_pack_arguments` purely to keep
+                    # the chain complete, and parking those would retain
+                    # full-size buffers that are never fetched.
+                    value = self._packed_arg_for_id.pop(idd)
+                    if idd in self._orphan_annotation_ids:
                         chain.append((prior_annotation, None, None))
+                    else:
+                        chain.append((prior_annotation, idd, value.target_dtype))
                 annotations_todo = chain + annotations_todo
             for annot, park_id, park_dtype in annotations_todo:
                 if annot.is_ext:
@@ -1286,6 +1294,7 @@ class CellComputationAutogradHooks(AutogradHooks):
                                 target_dtype=None,
                             )
                         )
+                        self._orphan_annotation_ids.add(self._next_id)
                         self._next_id += 1
         self._node_annotations.nodes.clear()
         # Flush all remaining pack arguments (these will not be packed)
--- a/keys_values/kvcache/gradient/autograd_hooks.py
+++ b/keys_values/kvcache/gradient/autograd_hooks.py
@@ -432,6 +432,16 @@ class CellComputationAutogradHooks(AutogradHooks):
     "scatter-key" annotation, which can then be done earlier (and skipped
     later).
 
+    More generally, the backward traversal may request an annotation whose
+    buffer state is more than one chunk behind the current final buffer. This
+    happens if autograd's backward for the intermediate chunks is served
+    without unpacking their "scatter-*" / "cat-*" annotations, e.g., when a
+    generated region spans three or more chunks (issue #148). In this case,
+    :meth:`_unpack_from_annotation` walks the annotation chain, applying
+    intermediate annotations early. Reconstructed intermediate states are
+    parked in `_id_to_unpacked`, so their IDs can still be served if autograd
+    asks for them later.
+
     Packing broadcast-extended indexes:
 
     A broadcast-extended index is a 4D tensor of integer dtype, obtained as
@@ -826,33 +836,65 @@ class CellComputationAutogradHooks(AutogradHooks):
                 print(
                     f"_unpack_from_annotation: {str(annotation)}, buffer={buffer.shape}, final_idx={final_idx}"
                 )
-            # This is complex, because it happens that "ext-*" appears before
-            # "scatter-*" or "cat-*" for the same node. In this case, we need
-            # to first execute this "prior annotation", since otherwise the
-            # input to "ext-*" does not exist.
-            annotations_todo = [annotation]
-            if annotation.is_ext:
-                if final_idx == chunk_idx + 1:
-                    # ext-* annotation comes too early, need to do another one first
-                    prior_annotation = self._find_prior_annotation(annotation)
+            # This is complex, for two reasons:
+            # (1) It happens that "ext-*" appears before "scatter-*" or
+            #     "cat-*" for the same node. In this case, we need to first
+            #     execute this "prior annotation", since otherwise the input
+            #     to "ext-*" does not exist.
+            # (2) The final buffer can be more than one chunk ahead of
+            #     `annotation`. This happens when autograd's backward for the
+            #     intermediate chunks is served without unpacking their
+            #     "scatter-*" / "cat-*" annotations (observed with generated
+            #     regions spanning 3+ chunks, see issue #148). All "scatter" /
+            #     "cat" annotations are kept in `_packed_arg_for_id` (see
+            #     :meth:`_flush_remaining_pack_arguments`), so we can walk the
+            #     chain here, applying intermediate annotations early. Their
+            #     reconstructed states are parked in `_id_to_unpacked`, in
+            #     case autograd asks for their IDs later on.
+            # `first_needed` is the buffer state (chunk index) required
+            # before `annotation` itself can be applied:
+            first_needed = chunk_idx if annotation.is_ext else chunk_idx + 1
+            # Entries are `(annot, park_id, park_dtype)`. If `park_id` is
+            # given, the state reconstructed by `annot` is parked under this
+            # ID (see (2) above)
+            annotations_todo = [(annotation, None, None)]
+            if final_idx < chunk_idx:
+                raise ValueError(
+                    f"Annotation {str(annotation)}: final chunk_idx = {final_idx}, must be >= {chunk_idx}"
+                )
+            elif not annotation.is_ext and final_idx == chunk_idx:
+                # Has already been done to support ext-* annotation
+                if self.debug_print_annotations:
+                    print("--> Skip (already done)")
+                annotations_todo = []
+            elif final_idx > first_needed:
+                # Walk the annotation chain from `final_idx - 1` down to
+                # `first_needed`, so that `annotation` can be applied last
+                chain = []
+                for prior_chunk_idx in range(final_idx - 1, first_needed - 1, -1):
+                    found = self._find_chain_annotation(annotation, prior_chunk_idx)
+                    if found is None:
+                        raise ValueError(
+                            f"Annotation {str(annotation)}: final chunk_idx = "
+                            f"{final_idx}: Cannot reconstruct, missing "
+                            f"'scatter'/'cat' annotation for chunk {prior_chunk_idx}"
+                        )
+                    idd, prior_annotation = found
                     if self.debug_print_annotations:
                         print(f"--> Doing {str(prior_annotation)} first")
-                    annotations_todo.insert(0, prior_annotation)
-                elif final_idx != chunk_idx:
-                    raise ValueError(
-                        f"Annotation {str(annotation)}: final chunk_idx = {final_idx}, must be in [{chunk_idx}, {chunk_idx + 1}]"
-                    )
-            else:
-                if final_idx == chunk_idx:
-                    # Has already been done to support ext-* annotation
-                    if self.debug_print_annotations:
-                        print("--> Skip (already done)")
-                    annotations_todo = []
-                elif final_idx != chunk_idx + 1:
-                    raise ValueError(
-                        f"Annotation {str(annotation)}: final chunk_idx = {final_idx}, must be in [{chunk_idx}, {chunk_idx + 1}]"
-                    )
-            for annot in annotations_todo:
+                    if prior_chunk_idx > chunk_idx:
+                        # Applied early: Remove the entry here, park the
+                        # reconstructed state below
+                        value = self._packed_arg_for_id.pop(idd)
+                        chain.append((prior_annotation, idd, value.target_dtype))
+                    else:
+                        # `prior_chunk_idx == chunk_idx` ("ext-*" case): The
+                        # entry stays in `_packed_arg_for_id`. If its ID is
+                        # unpacked later, the "already done" branch above
+                        # serves it
+                        chain.append((prior_annotation, None, None))
+                annotations_todo = chain + annotations_todo
+            for annot, park_id, park_dtype in annotations_todo:
                 if annot.is_ext:
                     # `target_dim1` is either `n_head` or `n_query_groups`,
                     # depending on whether the annotation includes extension
@@ -874,9 +916,22 @@ class CellComputationAutogradHooks(AutogradHooks):
                     self._node_annotations.set_final(
                         buffer,
                         layer_idx,
-                        chunk_idx,
+                        annot.chunk_idx,
                         kind,
                     )
+                    if park_id is not None:
+                        # `annot` was applied earlier than autograd asks for
+                        # it. Park the state just reconstructed (as a copy,
+                        # since `buffer` is modified in place by subsequent
+                        # steps), so :meth:`unpack_hook` can serve the ID later
+                        parked = buffer.clone()
+                        if park_dtype is not None and parked.dtype != park_dtype:
+                            parked = parked.to(dtype=park_dtype)
+                        self._id_to_unpacked[park_id] = parked
+                        if park_id not in self._id_counts:
+                            self._id_counts[park_id] = 1
+                        if self._debug_test_args:
+                            self._debug_log_args.append((parked.clone(), annot))
                 # Sanity check
                 assert (
                     annot.shape == buffer.shape
@@ -892,15 +947,30 @@ class CellComputationAutogradHooks(AutogradHooks):
     def _get_cache_length(self, layer_idx: int) -> int:
         return self.cache_lengths[layer_idx - self.first_layer_idx]
 
-    def _find_prior_annotation(self, annotation: NodeAnnotation) -> NodeAnnotation:
-        assert annotation.is_ext
+    def _find_chain_annotation(
+        self,
+        annotation: NodeAnnotation,
+        chunk_idx: int,
+    ) -> Optional[Tuple[int, NodeAnnotation]]:
+        """
+        Searches `_packed_arg_for_id` for the "scatter-*" or "cat-*"
+        annotation on the same chain as `annotation` (i.e., same layer and
+        keys/values kind), but with chunk index `chunk_idx`.
+
+        Args:
+            annotation: Annotation determining the chain
+            chunk_idx: Chunk index to search for
+
+        Returns:
+            `(id, chain_annotation)` if found, otherwise `None`
+
+        """
         layer_idx = annotation.layer_idx
-        chunk_idx = annotation.chunk_idx
         is_keys = annotation.is_keys
-        result = next(
+        return next(
             (
-                e.annot
-                for e in self._packed_arg_for_id.values()
+                (idd, e.annot)
+                for idd, e in self._packed_arg_for_id.items()
                 if (
                     isinstance(e, PackArgumentAsAnnotation)
                     and e.annot.is_keys == is_keys
@@ -911,11 +981,6 @@ class CellComputationAutogradHooks(AutogradHooks):
             ),
             None,
         )
-        if result is None:
-            raise IndexError(
-                f"{str(annotation)}: Don't find prior annotation for this one!"
-            )
-        return result
 
     @staticmethod
     def _unpack_scatter(
--- a/keys_values/kvcache/gradient/autograd_hooks.py
+++ b/keys_values/kvcache/gradient/autograd_hooks.py
@@ -203,6 +203,12 @@ class AnnotationUsageLog:
     unmatched_pack_args: List[UnmatchedPackHookArgument] = field(default_factory=list)
     unmatched_annotations: List[NodeAnnotationForLog] = field(default_factory=list)
     track_unmatched_annotations_for_pack_args: bool = False
+    # Peak memory retained (CPU) by parked buffer states within this cell,
+    # and the peak number of simultaneously parked states. Bounded by
+    # (num_chunks_per_cell - 1) * 2 * eff_num_layers buffers; typically far
+    # smaller, since states are freed as soon as their ID is fetched
+    parked_peak_bytes: int = 0
+    parked_peak_count: int = 0
 
     def report(self) -> str:
         lines: List[str] = [
@@ -211,6 +217,8 @@ class AnnotationUsageLog:
             f"Number of delta comparisons:             {self.num_comparisons}",
             f"Number of 4D indexes packed:             {self.num_4d_indexes}",
             f"Number of unmatched scatter annotations: {self.num_unmatched_scatter_cat}",
+            f"Parked states, peak (CPU):               {self.parked_peak_count} "
+            f"({self.parked_peak_bytes / (1 << 20):.1f} MiB)",
         ]
         if self.unmatched_annotations:
             lines.append("Remaining unmatched annotations:")
@@ -535,6 +543,10 @@ class CellComputationAutogradHooks(AutogradHooks):
         self._id_counts = None
         self._id_to_unpacked = None
         self._orphan_annotation_ids = None
+        self._parked_bytes_for_id = None
+        self._parked_bytes = 0
+        self._parked_peak_bytes = 0
+        self._parked_peak_count = 0
         # ID assigned to next pack hook argument
         self._next_id = None
         self._num_matched_annotations = None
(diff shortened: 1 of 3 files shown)


mmjerge:
@mseeger Fair questions — here is the full picture, including a concrete example. Short version: you are right that chunking is unaffected by generation, and right that there can be no holes in the reconstruction. Neither is violated. What breaks is the *order* in which autograd's backward requests the saved buffer states, and (in the GRPO path) which saved tensors get matched to annotations at all.

## What is different about our runs (it is not the data)

Two things, both specific to the RL path:

**1. The GRPO loop constructed the hooks without `may_match_twice` (config bug, fixed in this PR).** `finetune/longcontext_full.py` passes `may_match_twice=may_match_twice_factory(...)`, and with the new replay cache the `ext-*` annotations legitimately match *two* pack arguments. `keys_values/rl/grpo/loop.py` passed nothing, so the predicate defaulted to "never". Consequence: the second save of an ext buffer stays unmatched and is flushed as a raw tensor. Autograd can then run that chunk's SDPA backward entirely from the raw copy — the chunk's annotation is never unpacked at that point, so the final buffer does not advance down the chain. When an *earlier* chunk's annotation is later unpacked, the final buffer is now 2+ chunks ahead of it, and the old adjacency check `final_idx in [chunk_idx, chunk_idx+1]` raises. This is why you never see it: your path configures the predicate correctly.

**2. The backward traversal order deviates more in the GRPO graph.** The class docstring already documents that "the backward traversal ordering is not exactly the reverse of the forward traversal ordering", and the existing code special-cases exactly one deviation (an `ext-*` unpacked before its same-chunk `scatter-*`, handled via `prior_annotation`). With the GRPO head (per-token log-prob gather over the completion region + eos mask + ratio/clip, instead of the LM cross-entropy head), we observe larger deviations: after fixing (1), a 32k run still requested `scatter-value (27,15)` *after* the buffer had already been walked down to chunk 14 by an earlier request. Same phenomenon your `prior_annotation` mechanism handles, just at distance > 1 and in both directions. The PR generalizes that mechanism rather than adding a new one.

**How generation "plays a role":** it does not change chunking — chunk boundaries stay a function of cache length and chunk size only. But a gap of size g between the final buffer and a requested annotation needs g+1 saved buffer states between checkpoint columns whose unpack order deviates. With 32-token completions the loss-bearing region touches at most 2 chunks, so the reordering the GRPO graph induces has nothing to reorder across; with 2600-token completions it spans 3+ chunks. That is why hundreds of short-completion QA runs never hit it and long-generation runs hit it stochastically (the exact backward order also depends on which of the doubly-saved tensors matched, which is value-dependent).

## No holes — all deltas are applied, in order

The fix does not skip any deltas. Every `scatter`/`cat` annotation is already guaranteed to be present in `_packed_arg_for_id` when backward starts: matched ones under their pack-argument IDs, unmatched ones inserted by `_flush_remaining_pack_arguments` under fresh IDs — that is pre-existing behavior, added precisely so the chain is complete ("Even if a cat or scatter annotation is not matched to a pack argument, it may be required...").

When annotation (l, c) is requested with final buffer at f > c+1, the walk applies the deltas for chunks f-1, f-2, ..., c+1, then c — strict descending order, every link applied exactly once. If a link were genuinely missing, the new code raises naming the missing chunk (`Cannot reconstruct, missing 'scatter'/'cat' annotation for chunk k`); we have never seen that fire.

The one genuinely new piece of state: an intermediate annotation applied *early* has its reconstructed buffer **parked** (on CPU) under that annotation's ID. If autograd later asks for that ID — the "mirror" case above — it is served the exact state instead of hitting the walked-past buffer. The values are bit-exact: a scatter undo writes back the stored `delta` (the overwritten old values), so the parked state equals what a strictly-ordered unpack would have produced.

## Small example

This is literally what the two new unit tests construct (`test/kvcache/test_autograd_hooks.py`, runnable on CPU in seconds):

Take one layer, values only, cache length 32, three states linked by two scatters:

```
V1 --scatter(idx2, new2)--> V2 --scatter(idx3, new3)--> V3      final = V3 (chunk 3)
annotation (l,1): shape, idx2, delta2 = V1.gather(idx2)         # undoes chunk 2's scatter
annotation (l,2): shape, idx3, delta3 = V2.gather(idx3)         # undoes chunk 3's scatter
```

Backward asks for **chunk 1's state first** (its request for chunk 2's ID comes later, or is satisfied by a raw unmatched copy and never comes):

- **Old code:** `final_idx = 3, must be in [1, 2]` → `ValueError`. This is the exact failure from the issue logs (`scatter-value (20,1): final chunk_idx = 3`).
- **New code:** apply `(l,2)`: `V3.scatter_(idx3, delta3)` → V2, park a copy of V2 under `(l,2)`'s ID; apply `(l,1)`: → V1; return V1. Later request for `(l,2)`'s ID → served the parked V2 (second test: this request arrives after the buffer is already at V1, which the old code — and the first version of this PR — could not serve).

`test_unpack_walks_multi_chunk_annotation_chain` checks reconstruction for gaps of 1, 2, and 4 (gap 1 = unchanged legacy behavior); `test_unpack_out_of_order_request_after_walking_past` checks the parked-state serving. Both fail with the exact issue-#148 errors on the pre-PR code.

## Validation status (honest)

- The failures and both mechanisms are taken from real job logs (Qwen2.5-7B, L40S, `h2o-torch-quantized8`, 16k/8k cache, 2600-token and 32k-tier generations).
- An 8k long-generation GRPO run has since completed all 200 steps on the fixed code (previously crashed stochastically).
- 32k runs are in flight now; the first post-fix attempt OOMed because the parking initially cloned on-device — parked states now go to CPU.
- A `KV_DEBUG_ANNOTATIONS=1` traced 32k job is queued; I will attach the create/match/unpack sequence for the failing cell here when it lands, which will show the exact ordering deviation rather than my reconstruction of it.

mmjerge:
Traced run landed (`KV_DEBUG_ANNOTATIONS=1`, 32k-tier GRPO, Qwen2.5-7B, cache 16384, chunk 1024, layers_per_cell 1). Full log: `s3://keys-values-rl-results/runs/lm32k_trace/job.log`. It replaces my earlier reconstruction with observed behavior, and the observed behavior is simpler than what I described before.

## What the trace shows

Within each cell (4 chunks here), the backward's unpack requests arrive in strictly descending chunk order — no exotic reordering. Representative excerpt (layer 27, cell = chunks 21..24):

```
_unpack_from_annotation: ext-key   (27,24)  final_idx=24
_unpack_from_annotation: ext-key   (27,23)  final_idx=24   --> Doing scatter-key (27,23) first
_unpack_from_annotation: ext-value (27,23)  final_idx=24   --> Doing scatter-value (27,23) first
_unpack_from_annotation: ext-key   (27,22)  final_idx=23   --> Doing scatter-key (27,22) first
...
```

Two facts matter:

1. **The `prior_annotation` path fires for every chunk.** In this graph, the `ext-*` request for chunk c always arrives before the `scatter-*` ID for chunk c — so every scatter is applied early via your existing distance-1 mechanism. That happens ~40 times in this fragment alone. This is your design working as intended.

2. **The scatter IDs are then requested late — or the "on time" case never occurs at all.** There is not a single `--> Skip (already done)` line in 4k lines of trace. The skip branch assumes that when a scatter's own ID is unpacked, the final buffer still sits exactly at that chunk. In this graph the request arrives after later chunks' exts have already walked the buffer further down. That is precisely the failure from the real run: `scatter-value (27,15): final chunk_idx = 14, must be >= 15`. The buffer had legitimately moved to 14; the annotation asking for state 15 arrived afterwards.

So the invariant that breaks is narrower than "backward order deviates wildly": it is only the *lifetime* assumption behind the skip branch. The fix parks the state at the moment the scatter is applied early (one CPU copy per early-applied matched annotation, freed when served or at cell clear), so the late request is served bit-exactly.

The original gap-2 errors (`scatter-value (20,1): final chunk_idx = 3`) are the second variant: without `may_match_twice` (missing in the GRPO loop, fixed in this PR), the second save of an ext buffer is unmatched and served raw, so the ext→prior chain skips a link at that chunk, and the next chunk's ext arrives with a gap of 2. The chain walk covers that case.

## Why the RL path and not finetune ("why would a loss head do this")

I no longer believe the loss head is the mechanism. Inside a cell the graph structure is head-independent (the head enters only through the injected `head_gradients_top` values, and values don't affect engine ordering). The concrete differences in the RL path are:

- **Hooks config**: `may_match_twice` was missing (gap-2 variant above). Your paths always set it.
- **SDPA implementation**: the trace has 624 `padded-query` annotation events — these runs go through the zero-padded-query SDPA, not `flex_attention` (FlashInfer disabled on these boxes, and the qpadded fallback is what the RL example ends up on). Different SDPA = different consumers of the cache buffers = different backward request order for the scatter IDs. Your long-sequence runs are on the flex path.

That second point is a hypothesis I can test directly: run a finetune job under the same `KV_DEBUG_ANNOTATIONS=1` flag on the flex path and diff the request orderings. Happy to do that if it would settle it for you.

## On the efficiency invariants you raised

The change does not touch the matching side: annotations are still removed on first match (or marked once for `may_match_twice`), the flush is unchanged except for recording which IDs are orphans (a set of ints), and no pack decision changes — so no new unmatched pack args and no extra large nodes written into the graph. The walk adds work only when a request arrives out of order, and each delta is still applied exactly once overall (earlier, not additionally). Parked states go to CPU, not device.

## Remaining issue (separate from this PR)

The traced job — and the 32k reruns — now die in an eager-SDPA temp allocation (`sdpa_attention_weights` softmax, 2.31 GiB) with 40.9 GiB already allocated, i.e. a plain capacity problem in the qpadded path at cache 16384. The GRPO examples don't set `attention_forward_temp_size_gb` / backward temp limits the way `finetune_long_lora` does; we're fixing our recipe. Not an annotations problem.

mmjerge:
Ran the two discriminating experiments discussed above. Traces in `s3://keys-values-rl-results/runs/x148_*`.

## Experiment B: flex_attention path (your own `test_gradient_row_of_cells`, CUDA, h2o parametrizations, `debug_print_annotations=True`)

32 tests passed; trace has 1072 unpack requests. On the flex path your design assumptions hold exactly:

- 352 `prior_annotation` applications — **all distance 1** (the ext-before-scatter case you built the mechanism for; it is common, not rare).
- 336 direct scatter unpacks, **all on time** (`final_idx == chunk_idx + 1` in every single one).
- 576 second matches (`Matched again`) — `may_match_twice` is doing real work on this path.
- Zero adjacency/lifetime violations. This is consistent with these tests having always been green: **the invariant genuinely holds on the flex path.**

## The qpadded RL traces, for contrast

- 32k longmath GRPO (`lm32k_trace`): every chunk's scatter is applied via the ext-prior path; there is **not one** "on time" scatter unpack (`Skip (already done)`: 0 hits in 4k lines). The scatter IDs' own requests arrive after the buffer has moved on — that is the real-run failure `scatter-value (27,15): final chunk_idx = 14` and the reason the parking half of this PR exists.
- lp5 longproc GRPO (2600-token completions, cache 8192, `x148_mm_off_trace`): different again — scatters are requested **directly and on time** (final = chunk+1 throughout), and the ext IDs are never requested at all. So even within the qpadded path, the request ordering is configuration-dependent. This is, to me, the strongest argument for the generalized walk: the invariant is a property of a specific graph, not of the machinery, and three configurations already exhibit three different orderings.

## Experiment A: `may_match_twice` removed (attribution test) — inconclusive so far, being honest

A 2-step run on the lp5 config with the kwarg deliberately removed completed **cleanly** (trace shows 0 second matches, 0 unmatched, no gaps). So I have not yet demonstrated that the missing kwarg produces the original gap-2 errors — the original failures were stochastic (step ~10-60), and 2 steps is evidently not enough to hit the value-dependent conditions. A 60-step traced run without the kwarg is in flight; I will post the result either way. Two things remain true regardless of how it lands: the kwarg was missing from the GRPO loop while your paths and `test_gradient` set it (576 second matches in Experiment B show it matters on flex), and the chain walk + parking fix is required for the qpadded late-request behavior independent of the kwarg question.

One process suggestion from Experiment B: since `may_match_twice=None` silently degrades, would you take a follow-up that either makes the predicate required or logs a loud warning when a scatter/cat annotation goes unmatched while a same-shape pack argument went raw? That would have surfaced the GRPO config bug immediately.

mmjerge:
Update: we ran the discriminating experiment for the SDPA hypothesis, and the hypothesis is wrong. Correcting the record here.

We added flex support to the RL example (two omissions had kept it off: `flexatt_args` was never passed, and it must go into `cache_kwargs` because the gradient cells use `kv_cache.mha`, not `gpt_model.mha`). We then ran the identical 32k config, same seed, traced, on both paths:

```
                              qpadded          flex
unpack requests                  864             96 (partial; both OOMed later)
scatters applied early           432             72
  via ext-prior mechanism
direct scatter unpacks,            0              0
  on time
on-time skips                      0              0
second matches                   576             96
padded-query events              624              0  <- proves flex engaged
```

**Same ordering on both.** On our RL workload, flex also applies every scatter early through the `prior_annotation` path, with zero on-time scatter requests. So the SDPA variant is not what changes the backward ordering, and to answer your question directly: removing the qpadded variant would not remove the problem (and qpadded is also the only CPU path, so it cannot be removed anyway).

The corrected picture across all five traced setups:

- your `test_gradient_row_of_cells` (flex, tiny model): mixed mechanisms, every direct scatter request on time. Invariant holds.
- our longproc GRPO (qpadded, 7B): all direct, all on time. Invariant holds. 60 steps clean, including with `may_match_twice` removed — so we also cannot yet attribute the original gap-2 errors to the missing kwarg; that attribution stays unproven.
- our 32k longmath GRPO (qpadded AND flex, 7B): all scatters applied early, none on time, and the real run requested one after the buffer moved past it. Invariant broken.

Same machinery everywhere; the ordering is a property of the specific workload's graph, and we now have direct evidence it is not selected by the SDPA implementation. Which concrete property of the 32k graph causes autograd to consume the ext saves before the scatter saves, we have not isolated — but the measured claim the PR rests on is unchanged and now better supported: multiple orderings exist in practice, the current code assumes one, and the walk + parking make the reconstruction ordering-independent at zero cost to matching/packing (accounting in the earlier comment).

Separately: both 32k runs OOM later in the backward (identical 2.31 GiB allocation on both paths, with the forward attention temp limit set). That is a capacity/tuning issue in our 32k recipe, independent of the annotation machinery — the 8k and longproc configs train to completion on the fixed code.

mseeger:
Looking at this now. I need to really understand what is going on here.

mmjerge:
@mseeger Here is a reading order, and one honest note about your `ext-key (28,1494)` failure at the end.

## The mechanism in four sentences

`Annotations` keeps one live buffer per (layer, keys/values) plus its chunk index, `final`. Unpacking the scatter annotation for chunk `c` writes the stored old values back into that buffer, moving `final` from `c+1` to `c`. Your `_unpack_from_annotation` accepts a request for chunk `c` only if `final == c+1` (apply it) or `final == c` (already applied early by `prior_annotation`, return as is). That is an assumption about the order in which autograd asks for saved tensors: never more than one step ahead, never already past.

On long-generation GRPO with a 7B model, autograd asks in an order that violates both. Traced (third comment above): every scatter gets applied early through your `prior_annotation` path, and the scatters' own requests arrive after later chunks have walked the buffer further down. Same ordering on flex and on query-padded SDPA; the pre-PR code crashes on flex too.

## Reading order in the diff (`keys_values/kvcache/gradient/autograd_hooks.py` on this branch)

1. **Lines 904-918** — replaces the `[c, c+1]` check. `first_needed` is the state the request needs; if the buffer is *further ahead*, we walk instead of raising.
2. **Lines 918-944** — the walk. `_find_chain_annotation` (line 1027) finds the scatter/cat annotation for each intermediate chunk in `_packed_arg_for_id` — your flush already put every one of them there — and applies them in descending order. No delta is skipped; each is applied exactly once, just earlier than before.
3. **Lines 946-1005** — applying the list, and the parking (from 971). When a state is rebuilt earlier than autograd asks for it, one CPU copy is kept under that request's ID and freed when the request arrives. `ParkedBufferState` is line 153.
4. **Tests**, `test/kvcache/test_autograd_hooks.py`: `test_unpack_walks_multi_chunk_annotation_chain` (line 171) is the buffer-too-far-ahead case; `test_unpack_out_of_order_request_after_walking_past` (line 307) is the buffer-already-past case. Both fail on `main` with the exact production errors. `test_parked_memory_is_bounded_and_released` (line 444) pins the memory bound: parked states are freed on fetch, never created for chain links no autograd node references, and bounded by `(chunks_per_cell - 1)` per (layer, kind).

## Memory, measured

Qwen2.5-7B, cache 8192, 8k-token generations, L40S: device peak during backward 39.8 GiB with the fix on every one of 44 steps, 39.8 GiB with the pre-fix code on the same cache/generation configuration. Parked: 64 MiB, 2 states, on every step. Nothing new on the device; the walk mutates the same buffer the old code mutated.

## Your failure

```
ext-key (28,1494): final chunk_idx = 1493, must be in [1494, 1495]
```

Same disease, third symptom. Mine were (a) buffer too far ahead, (b) a scatter's own request arriving after the buffer moved past it. Yours is (c) an **ext** request arriving after the buffer moved past it. The parking in (3) covers (b) because it keys on the scatter's ID; it does not yet key on an outstanding ext for that chunk, so **this branch would still raise on your input** (line 909, with a `must be >= 1494` message). The extension is small and lives in the same block: when a scatter/cat state is produced and an `ext-*` for that chunk is still in `_packed_arg_for_id`, park the state for it; serve a late ext from the copy (`apply_ext_annotation` is pure, so this is safe). I will add it with a test that reproduces your error before we meet. The memory bound is unchanged: parking only ever happens for a consumer that is provably outstanding.

mseeger:
Hi Michael, I am looking at this. Maybe I get your explanations wrong.

But if `final_idx == chunk_idx + 2` in `_unpack_from_annotation`, this rather means that the annotation asks for a buffer too EARLY. The chain is at `final_idx`, and the request is for `chunk_idx`. From your comments, I read it that you see requests coming too LATE. Am I wrong?

My "ext" special treatment is for this early case. Namely, an "ext" annotation for `chunk_idx` can be requested when the chain is still at `chunk_idx+1`.

What needs to be done is to extend this to `chunk_idx < final_idx - 1` as well. Certainly the case `chunk_idx == final_idx - 2` will need to be covered. And maybe this should kick in whether `kind` is "ext" or something else.

I'll further dig into your change to understand whether this is what is going on.

In a sense, this is the simpler thing to fix. I was really worried that requests could come LATE. That would be real bad, because you'd almost have to store all of them somewhere, and you'd never know if a request could still come late.

If requests come early, we can simply run the chain and store the intermediates. Once they are requested, we can remove them. Right? It should be a rather simple extension of what I did already. Maybe this is what your fix is.

mseeger:
I still do not understand why this happens. I understand why it would ask for `"ext"(chunk_idx)` before `"scatter"(chunk_idx + 1 -> chunk_idx)`. But I do not understand why it would ask for `chunk_idx` before BOTH "scatter" `chunk_idx + 2 -> chunk_idx + 1` and `chunk_idx + 1 -> chunk_idx`.

This is beyond me. Why is the traversal of the computation graph so weird? I guess we have to accept that it sometimes is?

mseeger:
You must have spend lots of efforts on this. The fix seems clever, I need to fully understand it. The usage of `id_to_unpacked` is almost a bit "reckless", but my code is really not very clean and needs urgent refactoring at some point (there is an open issue for this).

I think your change will really make this thing more robust to other unexpected `autograd` reorderings.

mmjerge:
Yeah it took me a lot of time to fix it. It is a bit reckless, though, so I agree with you. Working on alternative methods, though it will work as it stands right now. I moved the parked states to their own dictionary with one helper for save, restore, and free, and _id_to_unpacked is reverted back to its original job. If you think its best to return back to the way I had it, let me know.

mmjerge:
I considered snapshotting every chunk, but that would require too much memory. I also thought of rebuilding states from checkpoints on demand, which would require a forward pass inside the unpack hook, but it would I think be too fragile. Unfortunately, there is no API for forcing autograd's ordering, though I wonder if we can add a feature in the library itself? So, keeping a copy only when something will provably ask for it was the cheapest safe option. Your ext-key crash is the same bug in a third form, and I'm extending the fix to cover it with a test that reproduces your exact error.

mseeger:
No, please leave it. AFAI can see, this is good as it is. I just need to fully understand it. Will work more on this tomorrow. To be fair, this takes quite some concentration.

mseeger:
This is also why I raised #31. All of this is getting a bit too messy.

mseeger:
Once I understand all this, I'll probably merge it. Can then tag further refactoring / cleanup ideas to #31. In the end, this fixes a real issue, and even makes the overall thing more robust.

mseeger:
As to your other comment: There is no way to affect `autograd` in any way. We need to just make sure that whatever happens there, this thing here works.

keys_values/kvcache/gradient/autograd_hooks.py
```diff
@@ -826,33 +909,97 @@ def _unpack_from_annotation(
...
+                    if self.debug_print_annotations:
+                        print(f"--> Doing {str(prior_annotation)} first")
+                    # The entry is applied early, so it is removed here. We
+                    # park the state it reconstructs (below), unless nothing
+                    # in the autograd graph can ask for this ID: orphan IDs
+                    # are inserted by
+                    # :meth:`_flush_remaining_pack_arguments` purely to keep
+                    # the chain complete, and parking those would retain
+                    # full-size buffers that are never fetched.
+                    value = self._packed_arg_for_id.pop(idd)
```
mseeger, line 1002: A bit lost here. Why can the entry be removed from `_packed_arg_for_id`?

You are probably right, but just explain it.

Normally, the entry is removed in `unpack_hook`. But indeed, as you run reconstructions early and move stuff into `_parked_states`, they do not come up in `unpack_hook` anymore.

But let us be sure about this, and also does not hurt to add a comment.

keys_values/kvcache/gradient/autograd_hooks.py
```diff
@@ -874,9 +1021,26 @@ def _unpack_from_annotation(
...
                         kind,
                     )
+                    if park_id is not None:
+                        # (2) `annot` was applied earlier than autograd asks
+                        #     for its own ID. Park a copy so :meth:`unpack_hook`
+                        #     can serve that ID later.
+                        self._parked_states[park_id] = self._park_buffer(
+                            buffer, park_dtype, ("id", park_id)
+                        )
+                        if park_id not in self._id_counts:
```
mseeger: What happens if `park_id in self._id_counts`? Would you not have to increase the counter?

But I am not fully behind this.

keys_values/kvcache/gradient/autograd_hooks.py
```diff
@@ -874,9 +1021,26 @@ def _unpack_from_annotation(
...
+                            self._id_counts[park_id] = 1
+                        if self._debug_test_args:
+                            self._debug_log_args.append((buffer.clone(), annot))
+                    # (3) If an "ext-*" annotation for this very chunk is still
+                    #     outstanding, autograd may ask for it after the buffer
+                    #     has moved on. Park a copy for it now.
+                    self._park_for_outstanding_ext(
```
mseeger, line 1056: What if `park_id is not None` here? Would the same buffer not be parked twice? If so, is this OK?

keys_values/kvcache/gradient/autograd_hooks.py
```diff
@@ -892,15 +1056,108 @@ def _unpack_from_annotation(
     def _get_cache_length(self, layer_idx: int) -> int:
         return self.cache_lengths[layer_idx - self.first_layer_idx]
 
-    def _find_prior_annotation(self, annotation: NodeAnnotation) -> NodeAnnotation:
-        assert annotation.is_ext
+    # --- Parked buffer states -----------------------------------------------
+    #
+    # A parked state is a copy of a reconstructed cache buffer that autograd
```
mseeger: I'd add that also autograd has asked for other nodes down the line, which is why the buffer was parked in the first place. Otherwise, one gets the impression that all nodes are parked. But we only park them because we have to, due to funny orderings.

keys_values/kvcache/gradient/autograd_hooks.py
```diff
@@ -892,15 +1056,108 @@ def _unpack_from_annotation(
...
+    # Copies live on CPU for device buffers. Every copy is released on fetch.
+
+    def _park_buffer(
+        self,
+        buffer: torch.Tensor,
+        target_dtype: Optional[torch.dtype],
+        key: Tuple,
+    ) -> Union[torch.Tensor, ParkedBufferState]:
+        dtype = target_dtype if target_dtype is not None else buffer.dtype
+        if buffer.device.type == "cpu":
```
mseeger, line 1100: Can we now avoid this, and have the entry be ParkedBufferState all the time?
We should avoid `Union` whenever possible.

keys_values/kvcache/gradient/autograd_hooks.py
```diff
@@ -892,15 +1056,108 @@ def _unpack_from_annotation(
...
+        self._parked_peak_count = max(
+            self._parked_peak_count, len(self._parked_bytes_for_key)
+        )
+        return parked
+
+    @staticmethod
+    def _restore_parked(
+        parked: Union[torch.Tensor, ParkedBufferState],
+    ) -> torch.Tensor:
+        if isinstance(parked, ParkedBufferState):
```
mseeger: I think you can avoid the `Union` here and elsewhere. Say the buffer is on CPU. Then, `.to` does nothing, no copy. I think this is cleaner.

keys_values/kvcache/gradient/autograd_hooks.py
```diff
@@ -892,15 +1056,108 @@ def _unpack_from_annotation(
...
+            for e in self._packed_arg_for_id.values()
+        )
+        if outstanding:
+            if self.debug_print_annotations:
+                print(
+                    f"--> Parking state ({layer_idx},{chunk_idx}) for outstanding ext"
+                )
+            self._ext_states[key] = self._park_buffer(buffer, None, ("ext",) + key)
+
+    def _find_chain_annotation(
```
mseeger, line 1157: Does this replace `_find_prior_annotation`?
What is the difference? We don't need both, right?

keys_values/kvcache/gradient/autograd_hooks.py
mmjerge: The entry has to be removed here because unpack_hook will never pop it for this ID. Autograd's request for this ID is now served from _parked_states instead, and that path does not touch _packed_arg_for_id. Leaving the entry in would retain the annotation and its delta tensors until clear(), and _find_chain_annotation could find the already-applied annotation a second time. I added a comment saying this (1e4f207).

keys_values/kvcache/gradient/autograd_hooks.py
mmjerge: Good catch, this line was wrong and I removed it (1e4f207). _id_counts must not be touched here. It counts how often autograd asks for an ID, and it only has entries for IDs matched at least twice. Parking does not change how often autograd asks, it only changes where the request is served from. Writing 1 into it also broke the invariant that entries are only for IDs matched twice or more. unpack_hook already treats a missing entry as one remaining request (_id_counts.get(idd, 1)), so no write is needed.

keys_values/kvcache/gradient/autograd_hooks.py
mmjerge: Yes, the same buffer state can be parked twice, once under ("id", park_id) and once under ("ext", layer, is_keys, chunk). This is intended. The two copies serve different requests, possibly with different dtypes, and each is released on its own when its request is served. Sharing one copy would complicate the release accounting for little gain, since this situation is rare. I added a comment (1e4f207).

keys_values/kvcache/gradient/autograd_hooks.py
mmjerge: Done (1e4f207). The header now says that parking only happens because autograd already asked for nodes further down the chain, and that nothing is ever parked in an in-order backward pass.

keys_values/kvcache/gradient/autograd_hooks.py
mmjerge: Done (1e4f207). _park_buffer now always returns ParkedBufferState, and the dicts are typed without Union. For a CPU buffer, device is the CPU device and restoring is a no-op.

keys_values/kvcache/gradient/autograd_hooks.py
mmjerge: Done (1e4f207). _restore_parked is now one line: parked.x_cpu.to(device=parked.device). There is no copy when the device is CPU, as you say. One detail: the parking side still needs an explicit clone for CPU buffers. A plain .to would alias the live buffer, and the live buffer is modified in place further down the chain. There is a comment at that spot.

keys_values/kvcache/gradient/autograd_hooks.py
mmjerge: Yes, it replaces _find_prior_annotation, which is gone from this branch. We only have the new one. Three differences. It takes the chunk index to search for, so the chain walk can look up any chunk on the chain, not just the annotation's own chunk. It also returns the pack-argument ID, which the chain walk needs in order to pop the entry and park the state under that ID. And it returns None instead of raising, so the caller can raise an error naming the exact missing chunk.

mmjerge:
I addressed all your comments in 1e4f207. The Union types are gone, the parked-state entries are always ParkedBufferState now, the redundant _id_counts write is removed, and the comments you asked for are added. All 18 tests pass.

One update on the ordering question. I filed pytorch/pytorch#198014 asking for the unpack ordering contract to be documented, and a docs PR pytorch/pytorch#198016. Edward Yang approved it and it merged into main. The PyTorch documentation now states that the unpack order of saved_tensors_hooks is unspecified and code must not depend on it. So there is no way to force an ordering, and none is planned. The defensive handling in this PR matches PyTorch's documented contract now, not just observed behavior.

On your refactoring ideas: I agree with both, and they belong in #31. Parking the outcome of ext annotations too would remove the second serving path and the special ext-state dictionary, at the cost of some extra parked bytes, which the accounting already measures. Moving the match and reconstruct logic to the caller would make autograd_hooks generic over annotation types. Happy to help with that once this is merged.

mseeger (approved):

